PR #1, opened 2016-04-12
Ek update pom repo

ekerwin:


akamen:
Looks good, just bummed that we integrations in one place and integration in another.  Nothing to do with this PR of course, but bit of an eye sore.

PR #5, opened 2016-04-18
Ek add password mgmt

ekerwin:
adding the encrypter/decrypter so that hub teamcity can consume them


src/main/java/com/blackducksoftware/integration/hub/encryption/PasswordDecrypter.java
```diff
@@ -0,0 +1,167 @@
...
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import org.slf4j.helpers.NOPLogger;
+
+public final class PasswordDecrypter {
+	private static Logger logger = LoggerFactory.getLogger(PasswordDecrypter.class);
+
+	private static final Charset UTF8 = Charset.forName("UTF-8");
+	private static final String EMBEDDED_SUN_KEY_FILE = "/Sun-Key.jceks";
+	private static final String EMBEDDED_IBM_KEY_FILE = "/IBM-Key.jceks";
```
psantos1113: Can we put some of the common contstants and methods that is in the Encrypter and Decrypter into a common base class.  Seems like the main difference between the 2 is the encrypt and decrypt methods.

jrichardBD: We may want to create a Util class that contains the constants and the shared methods Ex : getKey

We may also want to remove the Logger and pass in an IntLogger, then we can also remove the logError method since that should be handled by the Intlogger

psantos1113:
looks good

jrichardBD:
Looks good

PR #8, opened 2016-04-21
Updating validation

jrichardBD:


psantos1113:
looks good

ekerwin:
So much toString, equals and hashCode...sob. Otherwise, I can find no fault here. :)

PR #11, opened 2016-04-25
changing how proxy info is built

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/global/HubProxyInfo.java
```diff
@@ -18,67 +18,103 @@
...
+			final String ignoredProxyHosts) {
 		this.host = host;
 		this.port = port;
-		this.ignoredProxyHosts = noProxyHosts;
-		this.ignoredProxyHostPatterns = noProxyHostsPatterns;
-		this.proxyUsername = proxyUsername;
-		this.proxyPassword = proxyPassword;
-		this.hubUrlIgnored = hubUrlIgnored;
+		this.username = username;
+		this.password = password;
```
psantos1113: Should we add the encrypter class here like we use with the HubCredentials class to encrypt and decrypt the password as needed?

src/main/java/com/blackducksoftware/integration/hub/global/HubProxyInfo.java
```diff
@@ -18,67 +18,103 @@
...
 	}
 
-	public String getIgnoredProxyHosts() {
-		return ignoredProxyHosts;
+	public void setDefaultAuthenticator() {
+		if (username != null && password != null) {
+			Authenticator.setDefault(new Authenticator() {
+				@Override
+				public PasswordAuthentication getPasswordAuthentication() {
+					return new PasswordAuthentication(username, password.toCharArray());
```
psantos1113: If we encrypt the password in the constructor then this would need to decrypt the password since it is in use here.

src/main/java/com/blackducksoftware/integration/hub/validate/HubServerConfigValidator.java
```diff
@@ -0,0 +1,61 @@
...
+import com.blackducksoftware.integration.hub.exception.ValidationException;
+import com.blackducksoftware.integration.hub.job.HubScanJobConfigBuilder;
+import com.blackducksoftware.integration.hub.logging.IntExceptionLogger;
+
+public abstract class HubServerConfigValidator<T> {
+	public abstract T handleValidationException(ValidationException e);
+
+	public abstract T handleSuccess();
+
+	public T validateServerUrl(final String scanMemory) throws IOException {
```
psantos1113: This method name seems incorrect.  Looks like you're validating the scan memory and not the hub server url.

src/main/java/com/blackducksoftware/integration/hub/validate/HubServerConfigValidator.java
```diff
@@ -0,0 +1,61 @@
...
+package com.blackducksoftware.integration.hub.validate;
+
+import java.io.IOException;
+
+import com.blackducksoftware.integration.hub.ValidationExceptionEnum;
+import com.blackducksoftware.integration.hub.exception.ValidationException;
+import com.blackducksoftware.integration.hub.job.HubScanJobConfigBuilder;
+import com.blackducksoftware.integration.hub.logging.IntExceptionLogger;
+
+public abstract class HubServerConfigValidator<T> {
```
psantos1113: Will more validation methods for all the other fields in HubScanJobConfig be added here as well?

jrichardBD:
@ekerwin @psantos1113

jrichardBD:
@psantos1113

psantos1113:
@ekerwin @jrichardBD please review the changes to the validation.  I was able to remove the need for the ValidationException which can be an expensive operation and updated it to have only one abstract method to process a ValidationResult object into whatever result the CI needs.  I think Hub Jenkins will need to be updated to work with the changes here.

jrichardBD:
@ekerwin  @psantos1113  Can we merge this already? These changes are needed for Hub Jenkins master and I need them in the Atlassian refapp stuff

psantos1113:
@jrichardBD @ekerwin If we like the changes that are in the branch then we can merge.  I don't have any more functional changes to add.  The ValidationException and ValidationExceptionEnum class I don't think are used anymore so they should be removed but I didn't want to remove them if Hub Jenkins was compiling with them.

PR #13, opened 2016-05-17
Ekerwin updating test for windows

ekerwin:
@akamen @jrichardBD @psantos1113 @stevebillings 
These changes are a little more involved than simply updating the test, so I wanted to have them reviewed. It might be appropriate to have a sit-down code review here to make sure that I've correctly addressed all the important edge cases.


jrichardBD:
The changes look good as far as I can tell, but I wouldnt mind sitting down with you to go over these changes in case I am missing something

psantos1113:
I'd be interested in learning more about the code changes to the stream and what cases you ran into that caused the change to better understand it.  Otherwise I think the changes here are good as well.

PR #15, opened 2016-05-17
adding the logging code from James/Rob

ekerwin:


ekerwin:
@akamen @jrichardBD @psantos1113 @stevebillings @robpacheco 
I've moved Rob's and James's changes to a branch off of master.

PR #2, opened 2016-04-14
Add validators for the project and the version as separate methods

psantos1113:

PR #21, opened 2016-05-23
Ijrah 12 adding hub get methods

stevebillings:
The changes related to the JIRA integration are ready for review:

1. Split HubIntRestService in two: HubIntRestService is now Hub-focused. The lower-level Rest-focused methods have been moved to the new class: RestConnection. 
The lower-level Rest-focused methods that were part of the HubIntRestService API are still there, but are deprecated, and delegate to their replacements in RestConnection. 
To minimizes the chances of breaking existing functionality, I made minimal changes to pre-existing methods. There is an opportunity to modify some of the pre-existing methods to the new methods I added (see #2), but I have not done that. I also did not rename pre-existing http-get methods from get...() to httpGet...(). (I did make that change on the new methods.)

1. I added to RestService methods for getting resources from absolute and relative URLs, as well as some convenience methods.

1. I added the PolymorphicHubItemListService class, with methods for getting item lists from the Hub, where items can be of mixed type (subtypes of a common parent).


src/main/java/com/blackducksoftware/integration/hub/rest/RestConnection.java
```diff
@@ -0,0 +1,637 @@
...
+	    queryRef.addSegment(urlSegment);
+	}
+	for (AbstractMap.SimpleEntry<String, String> queryParameter : queryParameters) {
+	    queryRef.addQueryParameter(queryParameter.getKey(),
+		    queryParameter.getValue());
+	}
+	return queryRef;
+    }
+
+    // TODO not sure we want any static methods in this class
```
psantos1113: our convention for the most part has been to avoid having static methods unless it is a singleton or possibly a factory.  But we favor creating an instance of the object and using the reference.  If these are utility methods not needed to be a part of RestConnection then creating a utility class that is a singleton would work here.  That's just the convention our team has taken going forward with regards to static methods.

psantos1113:
The changes here look good you will need to merge the changes in master into this branch to resolve the conflicts.  Once this branch is merged I will need to merge master into the Updating_Validation before creating a pull request for that branch.  I have one question for clarification but it doesn't impact my thoughts on the pull request.  Looks good.

stevebillings:
@ekerwin @psantos1113 @jrichardBD @akamen Belatedly removing everyone's labels while I re-work this code.

stevebillings:
@ekerwin @psantos1113 @jrichardBD @akamen  OK, this is ready for another look. I've made the changes we discussed Monday:
- The new (http) get methods are now httpGet..., not get... I did not change existing methods.
- No more re-usable client resource.
- I've moved the lower-level (REST-specific, but not Hub-specific) methods from HubIntRestService to a new class: RestConnection. The lower-level methods that are part of the HubIntRestService API are deprecated, and delegate the work to RestConnection.
- Item is now HubItem
- Item lists are now parsed using the RuntimeTypeAdapterFactory approach.

I also changed pre-existing HubIntRestService methods to not call deprecated methods, but use the RestConnection methods instead. But that's the only type of change I made to pre-existing methods. There are opportunities to use the new httpGet* methods, but I did not make those changes since I wasn't eager to risk breaking pre-existing methods.

stevebillings:
@ekerwin @psantos1113 @jrichardBD @akamen Retracting this. Once the validator branch is merged into master, I'll merge that into this branch, then re-issue this pull request.

stevebillings:
@ekerwin @psantos1113 @jrichardBD @akamen  This is ready for review.

src/main/java/com/blackducksoftware/integration/hub/HubIntRestService.java
```diff
@@ -375,15 +253,15 @@ public ProjectItem getProject(final String projectUrl) throws IOException, BDRes
 
 	}
 
-	public ReleaseItem getProjectVersion(final String versionUrl)
-			throws IOException, BDRestException, URISyntaxException {
-		final ClientResource resource = createClientResource(versionUrl);
+	public ReleaseItem getProjectVersion(final String versionUrl) throws IOException, BDRestException,
+	URISyntaxException {
+		final ClientResource resource = getRestConnection().createClientResource(versionUrl);
```
psantos1113: We do a number of GET method calls.  I'm wondering if this should be encapsulated in RestConnection in a 
public ClientResource submitGETRequest(url) {
final ClientResource resource = restConnection.createClientResource(url);
		resource.setMethod(Method.GET);
		restConnection.handleRequest(resource);
 return resource;
}
we could add methods for PUT,POST, and DELETE as well to cover the verbs for REST calls.  Since we have the same pattern multiple times I was wondering if it makes sense to encapsulate it or because we may need ClientResource and RestConnection it is better to leave it the way it is and call the three lines of code in multiple places.

src/main/java/com/blackducksoftware/integration/hub/item/HubItemList.java
```diff
@@ -0,0 +1,37 @@
+package com.blackducksoftware.integration.hub.item;
+
+import java.util.List;
+
+/**
+ * A list of items as returned by the Hub.
+ *
+ * @author sbillings
+ *
+ */
```
ekerwin: it might be semantics but we could consider calling this HubItems but I'd avoid calling it a List unless it implemented java.util.List

src/main/java/com/blackducksoftware/integration/hub/item/PolymorphicHubItemListService.java
```diff
@@ -0,0 +1,123 @@
...
+import com.blackducksoftware.integration.hub.json.RuntimeTypeAdapterFactory;
+import com.blackducksoftware.integration.hub.rest.RestConnection;
+import com.google.gson.Gson;
+import com.google.gson.GsonBuilder;
+import com.google.gson.JsonArray;
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.reflect.TypeToken;
+
```
ekerwin: I'm not sure using 'polymorphic' here buys us much - it's just a list of T's, right?

stevebillings: Yes, it was originallycalled HubItemListService. The thought behind changing it to Polymorphic is that the list items' types must share a common parent. But I imagine all HubLists have that characteristic, so the "Polymorphic" in the name may be pointless.

src/main/java/com/blackducksoftware/integration/hub/item/HubItemList.java
stevebillings: Sure, that makes sense to me.

src/main/java/com/blackducksoftware/integration/hub/HubIntRestService.java
stevebillings: I agree that it makes sense to add methods to RestConnection for PUT, POST, DELETE. I was thinking it made sense to do that as the need for them arose.

Does RestConnection.httpGetFromAbsoluteUrl() not serve the purpose for submitting Get requests?

psantos1113: Sorry I missed that that is what I was talking about.  You already have it in place.

src/main/java/com/blackducksoftware/integration/hub/item/PolymorphicHubItemListService.java
stevebillings: During a conversation I just had with Rob it dawned on me that at a minimum, there will always be a common parent of HubItem, so, yeah, I'll drop the Polymorphic prefix, and we can use it for all lists.

stevebillings:
@psantos1113  @ekerwin @jrichardBD @akamen I've just incorporated Eric's feedback: renamed HubItems and HubItemsService classes.

jrichardBD:
Looks great to me!

PR #35, opened 2016-06-24
removing recorder references

ekerwin:
In line with the new plugins, we aren't going to want the CI's to create dependency lists anymore.


ekerwin:
@akamen @jrichardBD @psantos1113 @stevebillings

PR #36, opened 2016-07-22
Ps email notifier

psantos1113:
includes the notifications api objects used by the hub-jira project.  Also includes a fix to the post method of RestConnection


stevebillings:
Looks good

PR #42, opened 2016-08-16
Ps notification service refactor

psantos1113:
Changes to provide a dataservice to process notification items from the hub into usable objects.


psantos1113:
@stevebillings @jrichardBD @ekerwin please review.  We may need to add UUIDs to the content objects in order to use them in the hub-jira plugin.  Please take a look at these changes when you get a chance

stevebillings:
I like it.

PR #45, opened 2016-08-24
Ps performance improvements

psantos1113:
Changes to have Restlet use Apache HTTP Client.  Also ensure we release Response and ClientResource objects to allow the HTTP connection pool to recycle connections.


stevebillings:
The code looks good. A few weeks ago I overheard a conversation about a convention that package names be singular, but I don't know if that was something the team had agreed upon or one person's opinion. (It's not the kind of think I care about.)

jrichardBD:
Looks good to me @psantos1113

PR #46, opened 2016-08-31
Ps notification count digest

psantos1113:
Changes to get the count of notification data from the server rather than creating an object that represents the notification with all the relevant data in the object.  These changes are used by the email digest code


psantos1113:
@stevebillings @jrichardBD please review when you have a chance

psantos1113:
need to close this as the requirements are different so I need to make more code changes.

PR #49, opened 2016-09-13
adding code location api

ekerwin:


ekerwin:
I apologize that this pull request is so awful - the real important class is com.blackducksoftware.integration.hub.polling.ScanStatusService

ekerwin:
Please review at your leisure, @jrichardBD @psantos1113 @stevebillings @jessie-anderson @akamen

psantos1113:
@ekerwin I see all the constructors for the rest services now take the single and list item type token.  

super(restConnection, gson, jsonParser, new TypeToken<ReleaseItem>() {
		}.getType(), new TypeToken<List<ReleaseItem>>() {
		}.getType());

I understand how it works but for some reason it doesn't look neat to me.

Could we use some private final static members to initialize the type instances and pass them into the constructor.

private final static Type ITEM_TYPE  = new TypeToken<ReleaseItem>() {}.getType();
private final static Type LIST_TYPE = new TypeToken<List<ReleaseItem>>(){}.getType();

public ProjectVersionRestService(final RestConnection restConnection, final Gson gson,
 			final JsonParser jsonParser) {
		super(restConnection, gson, jsonParser, ITEM_TYPE, LIST_TYPE);
 	}
}

How do you think that would look?

src/main/java/com/blackducksoftware/integration/hub/api/scan/ScanStatusTypeEnum.java
```diff
@@ -0,0 +1,37 @@
...
+	COMPLETE,
+	CANCELLED,
+	CLONED,
+	ERROR_SCANNING,
+	ERROR_SAVING_SCAN_DATA,
+	ERROR_MATCHING,
+	ERROR_BUILDING_BOM,
+	ERROR;
+
+	private static final List<ScanStatusTypeEnum> PENDING_STATES = Arrays.asList(UNSTARTED, SCANNING, SAVING_SCAN_DATA,
```
psantos1113: Can this be a set so that the lookup in the contains below is constant time rather than iterating through the list?

src/main/java/com/blackducksoftware/integration/hub/api/scan/ScanStatusTypeEnum.java
```diff
@@ -0,0 +1,37 @@
...
+			SCAN_DATA_SAVE_COMPLETE, REQUESTED_MATCH_JOB, MATCHING, BOM_VERSION_CHECK, BUILDING_BOM);
+	private static final List<ScanStatusTypeEnum> DONE_STATES = Arrays.asList(COMPLETE, CANCELLED, CLONED,
```
psantos1113: same as above with using a set rather than a list.

src/main/java/com/blackducksoftware/integration/hub/polling/ScanStatusService.java
```diff
@@ -0,0 +1,169 @@
...
+				artifactId, version);
+		if (StringUtils.isBlank(mappedProjectVersionLink)) {
+			throw new UnexpectedHubResponseException(
+					String.format("Couldn't find a mapped project/version for: %s %s.", artifactId, version));
+		}
+
+		scanSummariesLinks = mappedProjectVersionToScanSummariesLinks.get(mappedProjectVersionLink);
+	}
+
+	private void waitUntilPendingScansAreComplete(final List<ScanSummaryItem> pendingScanSummaries,
```
psantos1113: Rather than recursive calls would using a FutureTask with the timeout set be possible.  You'd have to implement a Callable or runnable class to call waitUntilPendingScansAreComplete.  In waitUntilPendingScansAreComplete them implementation might be along the lines of:
`while(!pendingScanSummaries.isEmpty()) {

    Iterator<ScanSummaryItem> pendingItemIterator = pendingScanSummaries.iterator();  
    while(pendingItemIterator.hasNext()) {
        ScanSummaryItem item = pendingItemIterator.next();
        final String scanSummaryLink = item.getMeta().getHref();
	final ScanSummaryItem currentItem = scanSummaryRestService.getItem(scanSummaryLink);
        if (!currentItem.getStatus().isPending()) {
            pendingItemIterator.remove();
        }
    }
    Thread.sleep(FIVE_SECONDS);
}`

src/main/java/com/blackducksoftware/integration/hub/api/scan/ScanStatusTypeEnum.java
ekerwin: EnumSet coming soon to a different branch near you...

PR #53, opened 2016-09-16
Ps rule violation cleared notification

psantos1113:
Changes to support the policy violation cleared notification


psantos1113:
@ekerwin @jrichardBD @stevebillings  please review for support for the policy violation cleared notification.  Also updated the code to only wait for the notifications that were submitted for transformation and skip the others.  However with the way RuntimeTypeAdapterFactory works we may never run into that scenario and get an exception from GSon.  But the additional concern @stevebillings about hanging a plugin should be addressed now as well.

stevebillings:
Looks great.

PR #58, opened 2016-10-04
change to PhaseEnum that fixed a bamboo bug

jessie-anderson:


ekerwin:
This is absolutely correct, however, in the new world (coming soon!) I would recommend that we not decorate the objects meant to represent hub requests or responses. We might want that in a different place.

PR #60, opened 2016-10-05
Additional extension services

psantos1113:
@ekerwin please take a look.  Changes to use extension url rather than extension ID.


psantos1113:
spoke with reviewer in person and was approved.

PR #61, opened 2016-10-07
Replace ids with urls

stevebillings:
Here are the hub-common changes to move us away from parsing UUIDs from URLs, and assembling URLs from UUIDs. For those methods used by hub-jira, I wrote a replacement. For those not used by hub-jira, I added a TODO comment saying the replacement needs to be written. In both cases, I've deprecated the method.

There are some breaking changes because I removed UUID fields (replacing them with URL fields). I believe the breaking changes are limited to VulnerabilityRestService, NotificationContentItem and its subclasses, notification Transformers, ProjectVersion (plus maybe a handful methods that now throw an additional exception type).

To get the URL or relative URL for any HubItem, call item.getMeta().getHref() or .getRelativeHref().
To get the URL or reletive URL of a ProjectVersion, call getUrl() or getRelativeUrl().


stevebillings:
@ekerwin @psantos1113 Hi guys, could you review my description (above) at least, and the code too if you're interested? I'm done for today, but we can discuss Tuesday if you want.

stevebillings:
@psantos1113 Regarding Rob's comment above: I believe you are about to remove the aggregators that use VulnerabilityRestService.getComponentVersionVulnerabilities(), is that correct?

src/main/java/com/blackducksoftware/integration/hub/api/vulnerabilities/VulnerabilityRestService.java
```diff
@@ -71,6 +73,30 @@ public VulnerabilityRestService(final RestConnection restConnection, final Gson
...
 
+	public List<VulnerabilityItem> getComponentVersionVulnerabilities(final String componentVersionUrl)
+			throws IOException, URISyntaxException, BDRestException {
+		final HubRequest itemRequest = new HubRequest(getRestConnection(), getJsonParser());
+		itemRequest.setMethod(Method.GET);
+		itemRequest.setUrl(componentVersionUrl);
+		itemRequest.addUrlSegment(SEGMENT_VULNERABILITIES);
+
+		final JsonObject jsonObject = itemRequest.executeForResponseJson();
+		final List<VulnerabilityItem> allItems = getAll(jsonObject, itemRequest);
```
stevebillings: Thanks @robpacheco. Paulo is going to submit the bug.

PR #62, opened 2016-10-11
Ps vulnerability notification fixes

psantos1113:
@stevebillings  can you please review the changes when you get a chance


stevebillings:
Looks good. As it turns out, hub jira will use the VulnerableBomComponentRestService.getVulnerableComponentsMatchingComponentName() method that you added.

PR #65, opened 2016-10-13
Make Rest Connection abstract and add a CredentialsRestConnection

psantos1113:
Make the Rest Connection class abstract in order to remove the need for
a HubServerConfig object to be populated correctly to create the rest
connection.  Create a CredentialsRestConnection that will take the
HubServerConfig object in the constructor and initialize the
RestConnection object. If the username and password are empty then do
not call the setCookies method in the constructor.

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/rest/CredentialsRestConnection.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/rest/CredentialsRestConnection.java
@@ -1,0 +1,39 @@
+package com.blackducksoftware.integration.hub.rest;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+
+import org.apache.commons.lang3.StringUtils;
+import org.restlet.Context;
+import org.restlet.resource.ClientResource;
+
+import com.blackducksoftware.integration.exception.EncryptionException;
+import com.blackducksoftware.integration.hub.exception.BDRestException;
+import com.blackducksoftware.integration.hub.global.HubProxyInfo;
+import com.blackducksoftware.integration.hub.global.HubServerConfig;
+
+public class CredentialsRestConnection extends RestConnection {
+
+	public CredentialsRestConnection(final HubServerConfig hubServerConfig)
+			throws IllegalArgumentException, URISyntaxException, BDRestException, EncryptionException {
+		super();
+		setBaseUrl(hubServerConfig.getHubUrl().toString());
+		final HubProxyInfo proxyInfo = hubServerConfig.getProxyInfo();
+		if (proxyInfo.shouldUseProxyForUrl(hubServerConfig.getHubUrl())) {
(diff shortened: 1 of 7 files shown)


psantos1113:
@ekerwin @stevebillings @jrichardBD @jessie-anderson Please review the code where you use RestConnection you will need to change it to use CredentialsRestConnection.

src/main/java/com/blackducksoftware/integration/hub/rest/CredentialsRestConnection.java
```diff
@@ -0,0 +1,39 @@
...
+		super();
+		setBaseUrl(hubServerConfig.getHubUrl().toString());
+		final HubProxyInfo proxyInfo = hubServerConfig.getProxyInfo();
+		if (proxyInfo.shouldUseProxyForUrl(hubServerConfig.getHubUrl())) {
+			setProxyProperties(proxyInfo);
+		}
+		setTimeout(hubServerConfig.getTimeout());
+		final String userName = hubServerConfig.getGlobalCredentials().getUsername();
+		final String password = hubServerConfig.getGlobalCredentials().getEncryptedPassword();
+		if (StringUtils.isNotBlank(userName) && StringUtils.isNotBlank(password)) {
```
psantos1113: When the CredentialsRestConnection object is created we want to establish a connection to the Hub.  In all of the integrations we would get the HubServer config, create the rest connection, set the proxy properties, set the timeout, and then call set cookies.  First if you are creating this Credentials based RestConnection then you want to establish a session with the Hub.  So in talking to the team we decided to encapsulate that into the constructor.  Because we perform the same steps everytime.  If the constructor throws an exception we cannot establish a connection to the Hub at that time.  

The username I can definitely use rather than calling hubServerConfig.getGlobalCredentials().getUsername() again.  I will fix that there's no need for the multiple method calls.  I missed that.  The check on password is to determine only if you have a password set.  The setCookies method sets the values of j_username and j_password therefore for the password field we need to decrypt it before we can use it to set the j_password field.  If there is a problem decrypting the password then we cannot establish a connection to the hub and we will handle it appropriately depending on the integration.

robpacheco: It was really just the question of why not use the variables to set to the method instead of getting them again

stevebillings:
looks good to me.

src/main/java/com/blackducksoftware/integration/hub/rest/CredentialsRestConnection.java
ekerwin: We'd want to use the encrypted password to test for existence (it's the actual thing in the object) and use the decrypted password for connecting (since that can throw EncryptionException)

ekerwin:
Great stuff!

PR #70, opened 2016-11-03
Api user notification

psantos1113:
@stevebillings @ekerwin @jrichardBD 

Please review the changes for the user notifications so that we can release a new version of hub-common


psantos1113:
@stevebillings @ekerwin @jrichardBD  I uncovered an issue I need to look into further so I'm closing this pull request for now.  The format of user notifications isn't the same as the notifications base API.  Investigating it further.

PR #71, opened 2016-11-08
Api user notification

psantos1113:
@stevebillings @ekerwin @jrichardBD  Please review these are the changes for user notifications.


stevebillings:
Looks good.

ekerwin:
It'd be great if the date format stuff didn't leak out, but that's a fix we can address later.

ekerwin:
Love it.

psantos1113:
We'll have to add the start and end dates as members of the item request and encapsulate the date format and query parameters as something to go forward.

PR #73, opened 2016-11-10
component rest service implemented

jessie-anderson:


src/main/java/com/blackducksoftware/integration/hub/api/component/ComponentRestService.java
```diff
@@ -0,0 +1,48 @@
...
+import com.blackducksoftware.integration.hub.exception.BDRestException;
+import com.blackducksoftware.integration.hub.rest.RestConnection;
+import com.google.gson.Gson;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+import com.google.gson.reflect.TypeToken;
+
+public class ComponentRestService extends HubItemRestService<ComponentItem> {
+
+	private static final List<String> CODE_LOCATION_SEGMENTS = Arrays.asList(SEGMENT_API, SEGMENT_COMPONENTS);
```
robpacheco: Why is this called CODE_LOCATION_SEGMENTS? It looks like this is the component lookup api, right? something like /api/components?q=...

src/main/java/com/blackducksoftware/integration/hub/api/component/ComponentRestService.java
```diff
@@ -0,0 +1,48 @@
...
+	}.getType();
+
+	private static Type ITEM_LIST_TYPE = new TypeToken<List<ComponentItem>>() {
+	}.getType();
+
+	public ComponentRestService(final RestConnection restConnection, final Gson gson, final JsonParser jsonParser) {
+		super(restConnection, gson, jsonParser, ITEM_TYPE, ITEM_LIST_TYPE);
+	}
+
+	public List<ComponentItem> getAllComponents() throws IOException, BDRestException, URISyntaxException {
```
robpacheco: I could be mis-understanding this, (and perhaps this is also where i got lost above), but if this is a call to the component lookup api (external id lookup? -- I'm assuming that based on the name of this class), I'm pretty sure the 'q' parameter is required, but i don't see it specified here.

PR #77, opened 2016-11-18
fixing rest services and data services

jessie-anderson:


src/main/java/com/blackducksoftware/integration/hub/api/component/ComponentRestService.java
```diff
@@ -48,4 +50,20 @@ public ComponentRestService(final RestConnection restConnection, final Gson gson
         return allComponents;
     }
 
+    public ComponentItem getExactComponentMatch(String id, String groupId, String artifactId, String version)
```
robpacheco: one note here. The 'id' parameter, its not an actual id. It's the forge or namespace of the ID. A better name for this parameter would probably be forge (namespace is probably ok too).

ekerwin: We'd want to hold off applying any name changes until the Hub is updated to take a different value. Currently, a format that seems to be supported is "q=id:maven|org.apache.log4j|log4j|2.0.0" so keeping it as 'id' should be fine. If it should be something different, please make the change in the Hub API and documentation and our client will follow.

robpacheco: I think perhaps there is a mis-understanding. The id parameter is used here, as far as i can tell:

`final String idSegment = StringUtils.join(new String[] { "id", id }, ':');`

and in your sample, that will be 'maven'. 

Here, 'maven' is not an id, it is the forge type.

ekerwin: So the "id:" refers to the ENTIRE string, not just the first component of it. So forge, packageManager, namespace etc would cover all the "maven, rubygems, npmjs, nuget," strings. It's really just another piece OF the id the query wants to supplement the gav, right?

robpacheco: Yes, id: is the criteria to lookup by. the entire string after the: is the 'id', and is has up to 4 parts:
- forge/namespace
- group
- artifact
- version

(for maven/gradle)

These can be different is a different forge/namespace is used. For instance, npm i believe has only 2 parts in most cases.

PR #83, opened 2016-12-05
Adding RestConnection connect

jrichardBD:
IHCIC-76 Adding the connect method. Users of RestConnection will have to call connect now.


jrichardBD:
@stevebillings @ekerwin @psantos1113 Please review when you have a chance

PR #84, opened 2016-12-07
Validator implementation

psantos1113:
Changes to use the new Validators along with the builders to separate the work.


psantos1113:
@ekerwin @stevebillings @jrichardBD  please review along with the integration-common pull request.

PR #85, opened 2016-12-14
Creating meta service

jrichardBD:


jrichardBD:
The important stuff is really in the MetaService class if you guys just want to look at that

PR #88, opened 2016-12-28
Component license dataservice

ntrowles:


src/main/java/com/blackducksoftware/integration/hub/api/component/version/ComponentVersionRequestService.java
```diff
@@ -0,0 +1,114 @@
...
+import com.google.gson.JsonElement;
+import com.google.gson.JsonObject;
+
+public class ComponentVersionRequestService extends HubRequestService {
+
+	public ComponentVersionRequestService(RestConnection restConnection) {
+		super(restConnection);
+	}
+
+	public List<License> getAllLicenses(final String componentId, final String versionId) throws HubIntegrationException{
```
stevebillings: (As we discussed) we've been asked not to construct URLs using IDs

src/main/java/com/blackducksoftware/integration/hub/api/component/version/License.java
```diff
@@ -0,0 +1,112 @@
...
+		String[] urlElements = license.split("/");
+		return urlElements[urlElements.length-1];
+	}
+	
+	@Override
+	public String toString() {
+		return ReflectionToStringBuilder.toString(this, RecursiveToStringStyle.JSON_STYLE);
+	}
+	
+	//http://stackoverflow.com/questions/8180430/how-to-override-equals-method-in-java
```
stevebillings: Should either override both equals and hashcode, or neither.

PR #100, opened 2017-02-21
Simple license

rottebds:
Pull request to get Hub Eclipse everything it needs.


src/main/java/com/blackducksoftware/integration/hub/api/component/version/License.java
```diff
@@ -0,0 +1,59 @@
...
+	}
+
+	public String getName() {
+		return name;
+	}
+
+	public OwnershipEnum getOwnership() {
+		return ownership;
+	}
+
```
ekerwin: We shouldn't need to override toString, as the behavior you are after is in HubItem

src/main/java/com/blackducksoftware/integration/hub/api/component/version/ComplexLicenseItem.java
```diff
@@ -64,4 +67,8 @@ public ComplexLicenseType getType() {
         return type;
     }
```
ekerwin: We shouldn't need to override toString, as the behavior you are after is in HubItem

src/main/java/com/blackducksoftware/integration/hub/dataservice/model/ComplexLicenseModel.java
```diff
@@ -0,0 +1,75 @@
...
+package com.blackducksoftware.integration.hub.dataservice.model;
+
+import java.util.List;
+
+import org.apache.commons.lang3.builder.RecursiveToStringStyle;
+import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
+
+import com.blackducksoftware.integration.hub.api.component.version.ComplexLicenseItem;
+
+public class ComplexLicenseModel {
```
jrichardBD: If this class has MetaInformation then it should extend HubItem and should be called an 'Item' not a 'Model'
If it is indeed a model around HubItems then this object should not have MetaInformation to deal with

src/main/java/com/blackducksoftware/integration/hub/rest/CredentialsRestConnection.java
```diff
@@ -105,7 +105,7 @@ public void clientAuthenticate()
                         response = getClient().newCall(request).execute();
                         logResponseHeaders(response);
                         if (!response.isSuccessful()) {
-                            throw new HubIntegrationException(response.message());
+                            throw new HubIntegrationException(String.format("Connection Error: %1$s %2$s", response.code(), response.message()));
```
jrichardBD: The %1$s should be changed to %s to be consistent everywhere, unless you are specifically using the same segment multiple times or out of order

src/main/java/com/blackducksoftware/integration/hub/rest/RestConnection.java
```diff
@@ -262,8 +262,8 @@ private Response handleExecuteClientCall(final Request request, final int retryC
                     return handleExecuteClientCall(request, retryCount + 1);
                 } else {
                     throw new HubIntegrationException(
-                            "There was a problem trying to " + request.method() + " this item : " + request.url().uri().toString() + ". Error : "
-                                    + response.message());
+                            String.format("There was a problem trying to %1$s this item: %2$s. Error: %3$s %4$s",
```
jrichardBD: The %1$s should be changed to %s to be consistent everywhere, unless you are specifically using the same segment multiple times or out of order

src/main/java/com/blackducksoftware/integration/hub/api/component/version/ComplexLicenseItem.java
```diff
@@ -25,9 +25,6 @@
 
 import java.util.List;
 
-import org.apache.commons.lang3.builder.RecursiveToStringStyle;
-import org.apache.commons.lang3.builder.ReflectionToStringBuilder;
-
 import com.blackducksoftware.integration.hub.api.item.HubItem;
 
 public class ComplexLicenseItem extends HubItem {
```
ekerwin: the getter for licenses should probably just be called 'getLicenses' - in a world where these model classes are auto-created, the getters would all be named directly after the property they get

src/main/java/com/blackducksoftware/integration/hub/api/component/version/LicenseItem.java
```diff
@@ -23,37 +23,30 @@
...
-	}
+
+public class LicenseItem extends HubItem {
+    private final CodeSharingEnum codeSharing;
+
+    private final String name;
+
+    private final OwnershipEnum ownership;
+
+    public LicenseItem(CodeSharingEnum codeSharing, String name, OwnershipEnum ownership) {
```
ekerwin: I suspect this constructor is not needed. If it is only used in tests, let's put the code in the test. If it is used elsewhere, that might smell of a more sinister issue.

src/main/java/com/blackducksoftware/integration/hub/dataservice/license/ComplexLicenseParser.java
```diff
@@ -55,25 +55,6 @@ public String parse(ComplexLicenseItem complexLicense) {
                 if (i < complexLicense.getWrappedComplexLicenseItems().size()) licenseText.append(operator);
                 i++;
             }
-
```
ekerwin: I don't really object to this implementation, though I don't find it immediately obvious what it is doing. Perhaps we'll keep the api the same and revise it later. I do, however, object to a toString() invocation messing with object state. I'd suggest toString() be an idempotent operation.

src/main/java/com/blackducksoftware/integration/hub/rest/CredentialsRestConnection.java
```diff
@@ -105,7 +105,7 @@ public void clientAuthenticate()
                         response = getClient().newCall(request).execute();
                         logResponseHeaders(response);
                         if (!response.isSuccessful()) {
```
ekerwin: Though I approve of this change, it was nice to learn about the numbers and what they can do!

PR #104, opened 2017-03-08
name version guesser

ekerwin:
Not the best possible impl. but I think it gets 80% done. Thoughts? Is this okay?


jrichardBD:
It seems fine to me but where would this be used besides Artifactory? Does nexus have similiar naming conventions? Or are they the same since both are just stores for whatever artifacts you pump into them?

ekerwin:
@jrichardBD good question - I'd say this is a general solution for whenever you have only a filename. We may be able to use this as a last resort for hub-linux, if we are looking at filenames, or if we ever do a hub-filesystem where you ONLY have filenames. Do you think for now it should belong only in hub-artifactory until such a need arises? The packaging of that is a little nuanced, but if you think it has merit, we can discuss it.

PR #109, opened 2017-05-17
Add a Badge which links to the Maven Central page

romeara:
Links users to the maven central page, which has syntax for including as a dependency in a variety of build systems


ekerwin:
Adding more badges at this point is difficult to maintain consistency - is there a real need here?

PR #110, opened 2017-05-19
Validation updates

jrichardBD:
autoImportHttpsCertificates is false by default. The integrations will have to expose a way for Users to configure the import


src/main/java/com/blackducksoftware/integration/hub/builder/HubServerConfigBuilder.java
```diff
@@ -62,11 +68,46 @@
...
     @Override
+    public HubServerConfig build() throws IllegalStateException {
+        try {
+            return super.build();
+        } catch (final IllegalStateException e) {
+            if (e.getMessage().contains("SunCertPathBuilderException")) {
+                if (autoImportHttpsCertificates) {
+                    final HubCertificateHandler handler = new HubCertificateHandler(getLogger());
+                    try {
+                        if (StringUtils.isNotBlank(hubUrl)) {
```
psantos1113: Is it possible for the hub url to be empty at this point?  if we get the SunCertPathBuilderException we tried to connect to a hub url correct?

psantos1113: Spoke to @jrichardBD  this is defensive coding.  Not worth holding up the pull request over this.

jrichardBD: We can probably make some assumptions or we can add more defensive code. Not sure what the team wants to do

jrichardBD:
Still making changes, so I removed the reviewers that I could until I have everything finalized

jrichardBD:
Please re-review these changes

PR #126, opened 2017-09-01
Updating the Hub server url validation

jrichardBD:


src/main/groovy/com/blackducksoftware/integration/hub/validator/HubServerConfigValidator.java
```diff
@@ -149,107 +153,69 @@ public void validateHubUrl(final ValidationResults result) {
...
             hubURL.toURI();
         } catch (final MalformedURLException e) {
-            result.addResult(HubServerConfigFieldEnum.HUBURL,
-                    new ValidationResult(ValidationResultEnum.ERROR, ERROR_MSG_URL_NOT_VALID));
+            result.addResult(HubServerConfigFieldEnum.HUBURL, new ValidationResult(ValidationResultEnum.ERROR, ERROR_MSG_URL_NOT_VALID));
             return;
         } catch (final URISyntaxException e) {
-            result.addResult(HubServerConfigFieldEnum.HUBURL,
-                    new ValidationResult(ValidationResultEnum.ERROR, ERROR_MSG_URL_NOT_VALID));
+            result.addResult(HubServerConfigFieldEnum.HUBURL, new ValidationResult(ValidationResultEnum.ERROR, ERROR_MSG_URL_NOT_VALID));
```
rottebds: Any reason why we do the same thing twice for two different exceptions? Would it be better to do one catch clause?

PR #133, opened 2017-10-19
Added methods to retrieve the Hub response based on MediaType

msenmurugan:
Added methods to retrieve the Hub response based on MediaType


msenmurugan:
Paulo and James have approved the PR. Merging it into master.

PR #135, opened 2017-11-06
Vulnerability media type fix

psantos1113:


src/main/groovy/com/blackducksoftware/integration/hub/api/HubMediaTypes.java
```diff
@@ -0,0 +1,28 @@
...
+ * software distributed under the License is distributed on an
+ * "AS IS" BASIS, WITHOUT WARRANTIES OR CONDITIONS OF ANY
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.blackducksoftware.integration.hub.api;
+
+public class HubMediaTypes {
+    public final static String VULNERABILITY_REQUEST_SERVICE = "application/vnd.blackducksoftware.list-1+json";
```
robpacheco: I would add V1 into the name somewhere, since you'll likely need V2 in the near future.

src/main/groovy/com/blackducksoftware/integration/hub/api/vulnerability/VulnerabilityRequestService.java
```diff
@@ -36,7 +37,7 @@ public VulnerabilityRequestService(final RestConnection restConnection) {
     }
 
     public List<VulnerabilityView> getComponentVersionVulnerabilities(final String vulnerabilitiesUrl) throws IntegrationException {
```
robpacheco: Might want to think about what it would look like to support two versions of this API, a V1 and a V2 - since a lot of additional data will be available in V2 and then more in V3. 

Might also want to add an issue to add support for V2 (V3 will likely be available as of Hub 4.4).

psantos1113: Thanks for the feedback we have two internal tickets to track the changes we need.  One is to address the immediate issue which this code change is for.  We want to release a hotfix release 18.4.1 to fix the immediate concern for Hub 4.3 support when Hub 4.3 releases.  The other internal ticket we have is to address your concerns listed here longer term.

PR #140, opened 2017-11-29
User role

bamandel:
Enhanced the new UserDataService class to also get roles instead of just projects

### DIFF
--- a/src/main/java/com/blackducksoftware/integration/hub/dataservice/user/UserDataService.java
+++ b/src/main/java/com/blackducksoftware/integration/hub/dataservice/user/UserDataService.java
@@ -23,13 +23,11 @@
  */
 package com.blackducksoftware.integration.hub.dataservice.user;
 
-import java.util.ArrayList;
 import java.util.List;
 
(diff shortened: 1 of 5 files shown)


src/main/java/com/blackducksoftware/integration/hub/dataservice/user/UserDataService.java
```diff
@@ -71,4 +72,16 @@ public UserDataService(final RestConnection restConnection, final UserRequestSer
         return resolvedProjectViews;
     }
 
+    public List<RoleView> getRolesForUser(final String userName) throws IntegrationException {
+        final UserView user = userRequestService.getUserByUserName(userName);
+        return getRolesForUser(user);
+    }
+
+    public List<RoleView> getRolesForUser(final UserView userView) throws IntegrationException {
+        final String userRolesLink = metaService.getFirstLink(userView, MetaService.ROLES_LINK);
```
jamesrichard91: Can you move this code into the UserRequestService? We typically have the requests handled by the request services.

PR #142, opened 2017-12-01
Refactoring request services

jamesrichard91:


jamesrichard91:
Before this PR is merged we have to release hub-common-response

src/main/java/com/blackducksoftware/integration/hub/service/HubResponseService.java
```diff
@@ -50,12 +56,14 @@
     private final URL hubBaseUrl;
     private final JsonParser jsonParser;
     private final Gson gson;
+    protected final MetaService metaService;
 
-    public HubResponseService(final RestConnection restConnection) {
+    public HubResponseService(final RestConnection restConnection, final MetaService metaService) {
```
bamandel: Instead of requiring the MetaService to be passed for every single class, would we be able to create the metaService once we have the restConnection object?

PR #149, opened 2017-12-11
Service name refactor

rottebds:
Renaming a bunch of methods and classes to address technical debt. This PR should NOT be merged while there are other open PRs, due to the high chance of merge conflicts.


rottebds:
For anyone who is reviewing this PR, I have a few things that I'd like to guide your attention to.

HubVersionService.java is the only class with functional changes

There were a few naming changes that are not formulaic. They are:
MetaService -> MetaHandler
CLIDownloadService -> CLIDownloadUtility
SimpleScanService -> SimpleScanUtility
HubResponseItemManager -> HubViewTransformer
HubResponseItemsManager -> HubMultipleViewTransformer
HubResponseAllItemsManager -> HubMassViewTransformer

src/main/java/com/blackducksoftware/integration/hub/service/HubMassViewTransformer.java
```diff
@@ -0,0 +1,141 @@
...
+import com.blackducksoftware.integration.hub.exception.HubIntegrationException;
+import com.blackducksoftware.integration.hub.model.HubView;
+import com.blackducksoftware.integration.hub.request.HubPagedRequest;
+import com.blackducksoftware.integration.hub.request.HubRequestFactory;
+import com.google.gson.JsonObject;
+import com.google.gson.JsonParser;
+
+import okhttp3.Response;
+
+public class HubMassViewTransformer {
```
psantos1113: Not Sure if this is a better name.  But how about HubAllViewDataTransformer? HubMassViewTransformer seems rather vague but I don't know if my suggestion is a better name.

src/main/java/com/blackducksoftware/integration/hub/service/HubMultipleViewTransformer.java
```diff
@@ -40,40 +40,40 @@
 
 import okhttp3.Response;
 
-public class HubResponseItemsManager {
-    private final HubResponseItemManager hubResponseItemManager;
+public class HubMultipleViewTransformer {
```
psantos1113: For a class name change what about HubViewCollectionTransformer.  It will produce a collection of view objects.

ekerwin:
I don't think we want HubViewsCollectionTransformer and its brother - double plural isn't really needed. HubViewsTransformer or HubViewCollectionTransformer are both great...but HubViewsCollectionTransformer means you dealing with a collection of a collection.

PR #153, opened 2017-12-29
adding api key support

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/global/HubServerConfig.java
```diff
@@ -65,9 +75,12 @@ public void print(final IntLogger logger) {
         if (getHubUrl() != null) {
             logger.alwaysLog("--> Hub Server Url : " + getHubUrl());
         }
-        if (StringUtils.isNotBlank(getGlobalCredentials().getUsername())) {
+        if (getGlobalCredentials() != null && StringUtils.isNotBlank(getGlobalCredentials().getUsername())) {
             logger.alwaysLog("--> Hub User : " + getGlobalCredentials().getUsername());
         }
+        if (StringUtils.isNotBlank(apiKey)) {
+            logger.alwaysLog("--> Hub API Key : " + getApiKey());
```
stevebillings: Is it OK (secure) to log the api key?

src/main/java/com/blackducksoftware/integration/hub/validator/HubServerConfigValidator.java
```diff
@@ -129,11 +113,21 @@ public ValidationResults assertProxyValid() {
...
-    public ValidationResults assertCredentialsValid() {
+    public ValidationResults assertCredentialsOrApiKeyValid() {
         final CredentialsValidator credentialsBuilder = new CredentialsValidator();
         credentialsBuilder.setUsername(username);
         credentialsBuilder.setPassword(password);
-        return credentialsBuilder.assertValid();
+        final ValidationResults credentialsResults = credentialsBuilder.assertValid();
+        if (credentialsResults.hasErrors()) {
+            // you can specify either username/password OR apiKey
+            if (StringUtils.isNotBlank(apiKey)) {
```
gkillough: Should this supersede username/password? If so we can switch the validation order.

src/main/java/com/blackducksoftware/integration/hub/global/HubServerConfig.java
gkillough: We definitely shouldn't. Ryan left the following comment in hub-common-rest: 
"We don't log the headers here, as they contain a secret (the API key)" (in com.blackducksoftware.integration.hub.rest.ApiKeyRestConnection.clientAuthenticate()).

psantos1113: I wouldn't log the API key being used. If we want to log maybe just log that an API key is being used. If we see Hub User we know we are dealing with a credentials rest connection.  If we see "Hub API Key Used" then we know we are using the APIKey rest connection.

src/main/java/com/blackducksoftware/integration/hub/validator/HubServerConfigValidator.java
psantos1113: If we specify the check for API key it would potentially operate a faster if we are using the API key.  We wouldn't spend cycles validating the invalid credentials.  For that reason I would like the if (StringUtils.isNotBlank(apiKey) {

} else {
//credentials check....
//check for errors and add missing api key if errors found
}

src/main/java/com/blackducksoftware/integration/hub/validator/HubServerConfigValidator.java
```diff
@@ -129,11 +113,23 @@ public ValidationResults assertProxyValid() {
...
+    // you can specify either username/password OR apiKey
+    public ValidationResults assertCredentialsOrApiKeyValid() {
+        final ValidationResults validationResults = new ValidationResults();
+
+        if (StringUtils.isBlank(apiKey)) {
+            final CredentialsValidator credentialsBuilder = new CredentialsValidator();
+            credentialsBuilder.setUsername(username);
+            credentialsBuilder.setPassword(password);
+            final ValidationResults credentialsResults = credentialsBuilder.assertValid();
+            validationResults.addAllResults(credentialsResults.getResultMap());
```
gkillough: This line needs to move between line 129 and 130.

jamesrichard91: If it is moved then the if on line 127 would not work correctly

gkillough: Yeah, this comment is outdated. I had another comment below that has since been resolved.

PR #154, opened 2018-01-04
Added JSON response type to Vulnerability BOM component Api

msenmurugan:
Added JSON response type to Vulnerability BOM component Api


ekerwin:
This will be solved outside of the hub-common library.

PR #161, opened 2018-02-15
Parallel thread fix

psantos1113:


psantos1113:
The executor in use needs to have shutdown called to clean up the threads.  Otherwise when we instantiate a new service a new thread factory is created therefore new threads are created and the old threads are not interrupted.  So implement Closeable to shutdown the threads.

PR #168, opened 2018-03-21
using latest api

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/service/model/ComponentVersionPolicyViolationCount.java
```diff
@@ -46,8 +46,8 @@ public ComponentVersionPolicyViolationCount(final NameValuePairView nameValuePai
             name = PolicySeverityType.valueOf(nameValuePair.name);
         }
```
ekerwin: Not in love with this, but hub-common-api has (correctly) changed value from String to Object, and Gson converts it to a Double. Any other ideas?

jamesrichard91: Are you against something like this?
` if (nameValuePair.value != null) {
            value = ((Double) nameValuePair.value).intValue();
         }`

jamesrichard91: If your worried about the casting, them I like your approach for now

src/main/java/com/blackducksoftware/integration/hub/api/view/MetaHandler.java
```diff
@@ -40,7 +40,7 @@ public MetaHandler(final IntLogger logger) {
     }
 
     public boolean hasLink(final HubView view, final String linkKey) throws HubIntegrationException {
-        final ResourceMetadata meta = view.meta;
+        final ResourceMetadata meta = view._meta;
```
psantos1113: To remain consistent with the Hub you have removed the serialized name annotation, is that a correct assumption here?

src/main/java/com/blackducksoftware/integration/hub/service/model/ComponentVersionPolicyViolationCount.java
```diff
@@ -46,8 +46,8 @@ public ComponentVersionPolicyViolationCount(final NameValuePairView nameValuePai
...
 
-        if (StringUtils.isNumeric(nameValuePair.value)) {
-            value = Integer.valueOf(nameValuePair.value);
+        if (nameValuePair.value != null && NumberUtils.isCreatable(nameValuePair.value.toString())) {
```
psantos1113: can you extract nameValuePair into a variable before the if statement then use use that might make it a little easier to read.

Object countJsonObject = nameValuePair.value;
if(countJsonObject != null && NumberUtils.isCreatable(countJsonObject.toString()) {
    value = NumberUtils.createNumber(countJsonObject.toString()).intValue();
}

PR #173, opened 2018-04-16
Adding user notifications

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/notification/AbstractPolicyTransformer.java
```diff
@@ -122,7 +122,7 @@ protected PolicyRuleView getPolicyRule(final String ruleUrl) throws IntegrationE
      * In Hub versions prior to 3.2, the rule URLs contained in notifications are internal. To match the configured rule URLs, the "internal" segment of the URL from the notification must be removed. This is the workaround recommended by
      * Rob P. In Hub 3.2 on, these URLs will exclude the "internal" segment.
      */
-    //TODO ekerwin - can this now be removed as we are only supporting Hub versions >= 4.4.0?
+    // TODO ekerwin - can this now be removed as we are only supporting Hub versions >= 4.4.0?
```
psantos1113: We need to revisit the policy rule processing because the policy name is now present in the notification.  We no longer need to traverse the Policy Rule Url to get the name

PR #175, opened 2018-04-20
Remove old notifications

ekerwin:


jamesrichard91:
If none of these classes are needed anymore that is amazing but it does look like there are some conflicts with the branch

PR #177, opened 2018-04-24
Hub bucket parallel processing

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/notification/NotificationResults.java
```diff
@@ -24,35 +24,29 @@
...
+import com.blackducksoftware.integration.hub.api.view.CommonNotificationState;
+import com.blackducksoftware.integration.hub.service.bucket.HubBucket;
 
-    private final List<Exception> exceptions;
+public class NotificationResults {
+    private final List<CommonNotificationState> notificationContentItems;
+    private final HubBucket hubBucket;
 
-    public NotificationResults(final SortedSet<NotificationContentItem> notificationContentItems, final List<Exception> exceptions) {
+    public NotificationResults(final List<CommonNotificationState> notificationContentItems, final HubBucket hubBucket) {
```
gkillough: Can we make this a generic Collection instead of a List?

psantos1113: We certainly can.

src/main/java/com/blackducksoftware/integration/hub/service/NotificationService.java
```diff
@@ -171,32 +155,17 @@ public NotificationViewResults getAllNotificationViewResults(final Date startDat
...
-        } catch (final IOException ex) {
-            logger.debug("Error closing processor", ex);
-            exceptionList.add(ex);
-        } finally {
-            results = new NotificationResults(contentList, exceptionList);
-        }
+    private NotificationResults createNotificationResults(final List<CommonNotificationState> commonNotifications) throws IntegrationException {
+        final NotificationResults results;
+        final List<UriSingleResponse<? extends HubResponse>> uriResponseList = getAllLinks(commonNotifications);
+        final HubBucket bucket = hubBucketService.startTheBucket(uriResponseList);
```
gkillough: So the HubBucketService will be responsible for handling things in parallel depending on how it was instantiated?

psantos1113: yes @ekerwin and I discussed this yesterday. It gives the user of the service the option to control it.

gkillough: Ok cool. Just making sure I understood completely.

PR #180, opened 2018-04-27
updating to use optional

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/service/SignatureScannerService.java
```diff
@@ -123,11 +124,12 @@ private void printConfiguration(final HubScanConfig hubScanConfig, final Project
     }
 
     private void preScan(final HubServerConfig hubServerConfig, final HubScanConfig hubScanConfig, final ProjectRequest projectRequest) throws IntegrationException {
-        final String localHostName = HostnameHelper.getMyHostname();
-        logger.info("Running on machine : " + localHostName);
+        final Optional<String> localHostName = HostNameHelper.getMyHostName();
+        HostNameHelper.assertHostNamePopulated(localHostName);
```
gkillough: Doesn't this defeat the purpose of optional? 
Also, SonarQube is going to complain if you don't call isPresent() before the get() on line 132. I know it's done in assertHostNamePopulated, but it will still be flagged as a bug.

PR #182, opened 2018-04-28
mildly improving hostname helper

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/service/model/HostNameHelper.java
```diff
@@ -60,10 +61,10 @@ public static String getMyHostname() {
                 }
             } catch (final SocketException se) {
                 // ignore this
-                return hostName;
+                return Optional.ofNullable(hostName);
             }
         }
-        return hostName;
+        return Optional.ofNullable(hostName);
     }
```
rottebds: What is the reason for multiple return paths here? I know you didn't add them, but I question keeping them.

PR #186, opened 2018-05-07
Notification result store last date

psantos1113:


src/main/java/com/blackducksoftware/integration/hub/service/NotificationService.java
```diff
@@ -163,10 +165,18 @@ private NotificationResults createNotificationResults(final List<CommonNotificat
...
-            return notification1.getCreatedAt().compareTo(notification2.getCreatedAt());
-        }).collect(Collectors.toList());
-        results = new NotificationResults(contentList, bucket);
+
+        final Comparator<CommonNotificationState> dateComparator = Comparator.comparing(CommonNotificationState::getCreatedAt);
+        Collections.sort(commonNotifications, dateComparator.reversed());
+
+        final SimpleDateFormat sdf = new SimpleDateFormat(RestConstants.JSON_DATE_FORMAT);
+        sdf.setTimeZone(TimeZone.getTimeZone("UTC"));
+
```
stevebillings: I don't know the code around this well, so I could be mistaken, but it seems to me we want to be as sure as we can that generation of the date string is only done in one place (so that it's done consistently). Would the NotificationResult class not be that place?

PR #188, opened 2018-05-07
trying to improve the request factory

ekerwin:
Quite a bit going on here.

Renaming CIEnvironmentVariables, killing off any help for extensions, updating the latest integration libraries, replacing the hostname helper with a better api, removing joda-time, and trying to make RequestFactory as useful as possible.


src/main/java/com/blackducksoftware/integration/hub/service/model/HubQuery.java
```diff
@@ -0,0 +1,37 @@
...
+ * KIND, either express or implied. See the License for the
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+package com.blackducksoftware.integration.hub.service.model;
+
+public class HubQuery {
+    private final String q;
+
+    public HubQuery(final String parameter) {
```
gkillough: Is there a time when 'parameter' won't be a key-value pair? I think we should have an alternate constructor that takes two strings (key and value), pairs them, and sets 'parameter' with the pairing. Let the string building happen here.

PR #195, opened 2018-05-14
Ps rename get notification content details

psantos1113:


gkillough:
We might want to reevaluate constantly creating new lists. The state of the NotificationContent should never change, so we will always be creating the same data when this method is called.

psantos1113:
Will re-evaluate the creation of the notification details list in another branch and upcoming pull request.

PR #197, opened 2018-05-16
Ps notification detail refactor

psantos1113:
Move the creation of the content detail objects out of the notification content objects.


src/main/java/com/blackducksoftware/integration/hub/notification/NotificationViewResults.java
```diff
@@ -27,21 +27,19 @@
 import java.util.List;
 import java.util.Optional;
 
-import com.blackducksoftware.integration.hub.api.view.CommonNotificationState;
-
 public class NotificationViewResults {
```
psantos1113: We may be able to kill this class if we move the fields into NotificationResults

PR #200, opened 2018-05-18
Simple notification result

ekerwin:


src/main/java/com/blackducksoftware/integration/hub/notification/content/detail/NotificationContentDetailFactory.java
```diff
@@ -52,32 +53,33 @@ public NotificationContentDetailFactory(final Gson gson, final JsonParser jsonPa
     }
 
     public List<NotificationContentDetail> generateUserContentDetails(final NotificationUserView view) {
-        return generateContentDetails(view.contentType, view.createdAt, view.notificationState, view.type, view.json);
+        return generateContentDetails(view.contentType, view.createdAt, Optional.of(view.notificationState), view.type, view.json);
```
jamesrichard91: Is view.notificationState guaranteed to not be null? If it can be null, we should use Optional.ofNullable()

src/main/java/com/blackducksoftware/integration/hub/notification/content/detail/NotificationContentDetailFactory.java
```diff
@@ -108,14 +112,16 @@ public NotificationContentDetailFactory(final Gson gson, final JsonParser jsonPa
                     componentValue = componentVersionStatus.component;
                 }
                 details.add(NotificationContentDetail.createDetail(content, contentType,
-                        createdAt, type, NotificationContentDetail.CONTENT_KEY_GROUP_POLICY, notificationState, content.projectName, content.projectVersionName, content.projectVersion, componentVersionStatus.componentName,
-                        componentValue, componentVersionStatus.componentVersionName, componentVersionStatus.componentVersion, policyName, policyUri, null, componentVersionStatus.componentIssueLink, null));
+                        createdAt, type, NotificationContentDetail.CONTENT_KEY_GROUP_POLICY, optionalNotificationState, Optional.of(content.projectName), Optional.of(content.projectVersionName), Optional.of(content.projectVersion),
```
jamesrichard91: If any of these values can be null then we should use Optional.ofNullable() here

src/main/java/com/blackducksoftware/integration/hub/notification/NotificationDetailResult.java
```diff
@@ -0,0 +1,67 @@
...
+    public NotificationDetailResult(final NotificationContent notificationContent, final String contentType, final Date createdAt, final NotificationType type, final String notificationGroup, final String contentDetailKey,
+            final Optional<NotificationStateRequestStateType> notificationState, final List<NotificationContentDetail> simpleNotificationContentDetails) {
+        this.notificationContent = notificationContent;
+        this.contentType = contentType;
+        this.createdAt = createdAt;
+        this.type = type;
+        this.notificationGroup = notificationGroup;
+        this.contentDetailKey = contentDetailKey;
+        this.notificationState = notificationState;
+        this.simpleNotificationContentDetails = simpleNotificationContentDetails;
```
psantos1113: change the name to notificationContentDetails remove simple prefix

src/main/java/com/blackducksoftware/integration/hub/notification/content/detail/NotificationContentDetailFactory.java
ekerwin: In this case, we are dealing with a user notification which will always have a notification state.

src/main/java/com/blackducksoftware/integration/hub/notification/content/detail/NotificationContentDetailFactory.java
ekerwin: The code has been fundamentally altered from this state, but your point is valid, even though it no longer applies here.

PR #205, opened 2018-05-30
ComponentService remediating link

gkillough:


src/main/java/com/blackducksoftware/integration/hub/service/ComponentService.java
```diff
@@ -100,4 +106,16 @@ public ComponentSearchResultView getExactComponentMatch(final ExternalId externa
         throw new HubIntegrationException("Couldn't get a componentVersion url from the component matching " + externalId.createExternalId());
     }
 
+    // TODO deprecate when the REMEDIATING_LINK is included in ComponentVersionView
+    public RemediatingVersionView getRemediationInformation(final ComponentVersionView componentVersionView) throws IntegrationException {
+        final String href = hubService.getHref(componentVersionView);
+        final String remediatingURL = href + "/" + REMEDIATING_LINK;
+        try (final Response response = hubService.executeGetRequest(remediatingURL);) {
+            final JsonElement jsonElement = hubService.getJsonParser().parse(response.getContentString());
```
gkillough: RemediatingVersionView is not considered a HubResponse in hub-common-api. It probably should be.

src/main/java/com/blackducksoftware/integration/hub/service/ComponentService.java
```diff
@@ -100,4 +106,16 @@ public ComponentSearchResultView getExactComponentMatch(final ExternalId externa
         throw new HubIntegrationException("Couldn't get a componentVersion url from the component matching " + externalId.createExternalId());
     }
 
+    // TODO deprecate when the REMEDIATING_LINK is included in ComponentVersionView
+    public RemediatingVersionView getRemediationInformation(final ComponentVersionView componentVersionView) throws IntegrationException {
+        final String href = hubService.getHref(componentVersionView);
+        final String remediatingURL = href + "/" + REMEDIATING_LINK;
```
gkillough: I feel like this is an example _for_ the separation of the URI/URL construction. Maybe _DataService_ and _HubService_ should extend something like _AbstractRestService_ which has that common functionality?

PR #208, opened 2018-06-18
Updating the parsing of the CLI logs

jamesrichard91:


src/main/java/com/blackducksoftware/integration/hub/service/model/ScannerSplitStream.java
```diff
@@ -220,36 +221,42 @@ public void close() throws IOException {
...
+            logger.error(trimmedLine);
+        } else if (StringUtils.containsIgnoreCase(trimmedLine, FINISHED)) {
+            outputBuilder.append(trimmedLine);
             outputBuilder.append(System.getProperty("line.separator"));
-            logger.info(line);
+            logger.info(trimmedLine);
         }
 
         output = outputBuilder.toString();
     }
```
ekerwin: I wonder if:
return line.toUpperCase().startsWith(string.toUpperCase())
might be better. You can even skip the second toUpperCase since you only call the method with all-caps strings.

src/main/java/com/blackducksoftware/integration/hub/service/model/ScannerSplitStream.java
```diff
@@ -220,36 +221,42 @@ public void close() throws IOException {
...
+        if (doesStartWith(trimmedLine, DEBUG) || doesStartWith(trimmedLine, TRACE)) {
             // We dont want to print Debug or Trace logs to the logger
             return;
         }
         final StringBuilder outputBuilder = new StringBuilder();
         outputBuilder.append(output);
-        if (StringUtils.containsIgnoreCase(line, EXCEPTION)) {
-            // looking for 'Exception in thread' type messages
-            outputBuilder.append(line);
+        if (doesStartWith(trimmedLine, ERROR)) {
```
ekerwin: It looks like some effort was done to turn strings into reasonable variables (ERROR, WARN, EXCEPTION, FINISHED, etc) so I see no reason to not also create LINE_SEPARATOR to reduce the noise in this method. You can replace all 'System.getProperty("line.separator")' occurrences with the variable. Seems cleaner to me.

PR #218, opened 2018-08-08
Updating to use the new phone home classes

jamesrichard91:


gkillough:
It doesn't look like we ever shutdown the ExecutorService in the phone-home-client.

PR #227, opened 2018-09-17
Use bds java home

ekerwin:


ekerwin:
@joc0611 I built off your PR and removed the need for powermock in the tests. We've tried (though not very successfully at times) in keeping dependencies on System (for the challenge in testing among other reasons) to a minimum. I think this will still fix your issues.

joc0611:
@ekerwin Perfect, thanks for your help and rework! Your argument makes sense. I thought about that solution as well, however I noticed that hub-detect also uses ``ScanPathUtility``, so I tried to avoid changing the constructor.

joc0611:
But I guess you can judge this way better than I can :-)

ekerwin:
@joc0611 Actually, the team has been discussing things after your changes were submitted and I think we're going to start using powermock. Thank you for such great contributions!

PR #230, opened 2018-10-03
Fix Java Executable Path for Signature Scanner

### LINKED ISSUE #224: Running Signature Scanner from Hub Detect on Alpine Linux fails due to Oracle JVM
I'm currently trying to run Hub Detect version 4.2.1 inside an ``openjdk:8-jre-alpine`` docker container. My Hub installation, and hence the ``scan.cli.zip``, is version 4.5.0. 

The ``scan.cli.zip`` is packaged with an Oracle JRE, which is not compatible with Alpine Linux (see [Alpine Linux Caveats](https://github.com/gliderlabs/docker-alpine/blob/master/docs/caveats.md#incompatible-binaries). For this reason, the signature scanner is failing with 

```
There was a problem scanning target '/home/jenkins/webapp-backend' : Exception executing the cli scan: Cannot run program "/home/jenkins/workspace/Blackduck/WebApp/tools/Hub_Scan_Installation/scan.cli-4.5.0/jre/bin/java": error=2, No such file or directory
```

Running the signature scanner from it's own ``/home/jenkins/workspace/Blackduck/WebApp/tools/Hub_Scan_Installation/scan.cli-4.5.0/bin/scan.cli.sh`` works fine, if the environment variable ``BDS_JAVA_HOME`` is set to the default OpenJDK JRE of the docker image.

However, Hub Detect 4.2.1 (through hub-common-37.0.0) seems to invoke the signature scanner directly from their Java code, ignoring the ``BDS_JAVA_HOME`` variable. I think the correct place for taking ``BDS_JAVA_HOME`` into consideration would be ``com.synopsys.integration.blackduck.cli.SimpleScanUtility#setupAndExecuteScan(com.synopsys.integration.blackduck.cli.CLILocation)``

Is my observation correct? Are there any known workarounds to invoke signature scanning from Hub Detect using the systems default JRE?

Any help is appreciated. Thanks!

joc0611:
When BDS_JAVA_HOME was set, its value was used as the java executable
path. By convention any JAVA_HOME variable usually points to the
installation directory of the JVM though, e.g.
/usr/lib/jvm/java-8-openjdk-amd64/jre. Hence, to determine the path to
the java executable correctly, we need to append /bin/java or
/bin/java.exe

Fixes #224

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/signaturescanner/command/ScanPathsUtility.java
+++ b/src/main/java/com/synopsys/integration/blackduck/signaturescanner/command/ScanPathsUtility.java
@@ -97,7 +97,7 @@ public class ScanPathsUtility {
         final String pathToJavaExecutable;
         final String bdsJavaHome = intEnvironmentVariables.getValue(BDS_JAVA_HOME);
         if (StringUtils.isNotBlank(bdsJavaHome)) {
-            pathToJavaExecutable = bdsJavaHome;
+            pathToJavaExecutable = findPathToJavaExe(new File(bdsJavaHome));
         } else {
             final File jreContentsDirectory = findFirstFilteredFile(installDirectory, JRE_DIRECTORY_FILTER, "Could not find the 'jre' directory in %s.");
             pathToJavaExecutable = findPathToJavaExe(jreContentsDirectory);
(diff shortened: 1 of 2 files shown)


joc0611:
Hey @ekerwin, do you know, how I could fix this:
```
* What went wrong:
Execution failed for task ':sonarqube'.
> You're only authorized to execute a local (preview) SonarQube analysis without pushing the results to the SonarQube server. Please contact your SonarQube administrator.
```

ekerwin:
@joc0611 No idea about the sonarqube thing - it isn't super important to me that it works. :) But your code here is great, as usual. Thanks for the help!

PR #231, opened 2018-10-03
Ek use env for targets

ekerwin:


src/main/java/com/synopsys/integration/blackduck/signaturescanner/command/ScanPaths.java
```diff
@@ -36,6 +39,26 @@ public ScanPaths(final String pathToJavaExecutable, final String pathToOneJar, f
...
+    public List<String> createCommandForScannerVersion() {
+        final List<String> cmd = new ArrayList<>();
+        addScanCommandPrefix(cmd);
+        addScanCommandSuffix(cmd);
+        cmd.add("--version");
+
+        return cmd;
+    }
+
+    public void addScanCommandPrefix(final List<String> cmd) {
```
taikuukaits: I don't love the name of these methods. I assumed prefix meant adding something to the prefix of each command sent in. I don't have a great alternative, but something like addStandardCommandOptions.

PR #234, opened 2018-10-10
using blackduck properties

ekerwin:


src/main/java/com/synopsys/integration/blackduck/configuration/HubServerConfigBuilder.java
```diff
@@ -331,39 +333,73 @@ public void setTrustCert(final boolean trustCert) {
...
+            return keys.contains(hubEnvironmentVariableKey) || keys.contains(hubPropertyKey) || keys.contains(blackDuckEnvironmentVariableKey) || keys.contains(blackDuckPropertyKey);
         }
 
         public String getValueFrom(final Map<String, String> values) {
-            if (values.containsKey(propertyKey)) {
-                return values.get(propertyKey);
-            } else {
-                return values.get(environmentVariableKey);
+            String key = hubEnvironmentVariableKey;
+            if (values.containsKey(hubPropertyKey)) {
```
jamesrichard91: Why are we preferring the hub property over the blackduck property?

PR #242, opened 2018-11-28
Added a simple upload service.

taikuukaits:
Here is an example Simple Upload Service that I (Detect) would want to use.

I am still working on the tests but wanted some initial feedback.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/bdioupload/SimpleUploadService.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/bdioupload/SimpleUploadService.java
@@ -1,0 +1,63 @@
+package com.synopsys.integration.blackduck.codelocation.bdioupload;
+
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.IOException;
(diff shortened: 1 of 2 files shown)


src/main/java/com/synopsys/integration/blackduck/codelocation/bdioupload/SimpleUploadService.java
```diff
@@ -0,0 +1,63 @@
...
+            codeLocationNameMap.put(codeLocationName, bdioFile);
+        }
+        return uploadNameMappedFiles(codeLocationNameMap);
+    }
+
+    private String readCodeLocationName(File bdioFile) throws IOException {
+        try (FileInputStream fileInputStream = new FileInputStream(bdioFile)){
+            try (BdioReader bdioReader = new BdioReader(new Gson(), fileInputStream)){
+                SimpleBdioDocument simpleBdioDocument = bdioReader.readSimpleBdioDocument();
+                String codeLocationName = simpleBdioDocument.billOfMaterials.spdxName;
```
jamesrichard91: What do we do if there is no billOfMaterials or spdxName? I know Detect will also set those, but what about other users?

PR #256, opened 2019-04-09
updating risk report component links

quarksandthings:
Component links in pdf risk report are not valid starting with 2018 backend.  
Project and project version links still work in 2019.06 backend.


ekerwin:
This is definitely an issue - thank you for bringing it to our attention! I think we'll address this with a more in-depth update since it appears we can just use the response from VersionBomComponentView directly instead of parsing its value and prepending a base url. I'll try to tackle this tomorrow and I'll add you as a reviewer - thanks again!

ekerwin:
Here's the commit that fixes the issue - you can plainly see that not parsing is better when we can get away with it! Thanks!
https://github.com/blackducksoftware/blackduck-common/commit/56247e2295b2301f9c224f56ce235d91459fc4ce

PR #262, opened 2019-06-17
Updating the BucketService to return Futures

jamesrichard91:


src/main/java/com/synopsys/integration/blackduck/service/bucket/BlackDuckBucket.java
```diff
@@ -88,6 +88,14 @@ public void addError(final String uri, final Exception e) {
         bucket.put(uri, new BlackDuckBucketItem<>(uri, e));
     }
 
+    public Boolean hasAnyErrors() {
+        return bucket.values()
+                   .stream()
+                   .filter(blackDuckResponseBlackDuckBucketItem -> blackDuckResponseBlackDuckBucketItem.hasException())
```
gkillough: Could this be a method reference?

jamesrichard91:
@gkillough Im not sure what you mean by that? The BlackDuckBucketFillTask does exactly that. Are you saying we should have an alternative that does that directly without the executor service?

gkillough:
> @gkillough Im not sure what you mean by that? The BlackDuckBucketFillTask does exactly that. Are you saying we should have an alternative that does that directly without the executor service?

I'm talking about a simpler wrapper around the bucket. The BucketFillTask always seemed a little convoluted to me. All I would want is a simple wrapper that either gets something from the bucket or invokes the BlackDuckService; let me, the user, handle the exception or maybe have a "safe" method in addition to this. Perhaps a discussion for the future.

src/main/java/com/synopsys/integration/blackduck/service/bucket/BlackDuckBucketFillTask.java
```diff
@@ -38,16 +41,19 @@ public BlackDuckBucketFillTask(final BlackDuckService blackDuckService, final Bl
     }
 
     @Override
-    public void run() {
+    public Optional<? extends BlackDuckResponse> call() {
```
ekerwin: as this represents a breaking api change, we'd also want to update the build.gradle to reflect the new semantic version

src/main/java/com/synopsys/integration/blackduck/service/bucket/BlackDuckBucketService.java
```diff
@@ -59,28 +61,30 @@ public BlackDuckBucket startTheBucket(final List<UriSingleResponse<? extends Bla
         return blackDuckBucket;
     }
 
-    public <T extends BlackDuckResponse> void addToTheBucket(final BlackDuckBucket blackDuckBucket, final String uri, final Class<T> responseClass) throws IntegrationException {
-        final List<UriSingleResponse<? extends BlackDuckResponse>> uriSingleResponses = new ArrayList<>();
-        uriSingleResponses.add(new UriSingleResponse<>(uri, responseClass));
-        addToTheBucket(blackDuckBucket, uriSingleResponses);
+    public Future<Optional<? extends BlackDuckResponse>> addToTheBucket(final BlackDuckBucket blackDuckBucket, final String uri, final Class<? extends BlackDuckResponse> responseClass) throws IntegrationException {
```
ekerwin: Moving to ? from T - is that lossy at all? Is the binding of the returned class the same as the method parameter? I thought each ? got its own binding.

PR #271, opened 2019-10-21
feat(service): Allows responses to be paged to a limit.

JakeMathews:
Completes INTCMN-416.
Allows the user to specify how many pages they would like to retrieve. I considered total count, but that got complicated quickly. If the user wants a more precise number, they can use the BlackDuckResponsesTransformer directly and pass in a custom PagedRequest with their own offsets and limits. BlackDuckResponsesTransformer is now exposed on BlackDuckService for convenience.

Also added direct tests for BlackDuckResponsesTransformer.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/service/BlackDuckResponsesTransformer.java
+++ b/src/main/java/com/synopsys/integration/blackduck/service/BlackDuckResponsesTransformer.java
@@ -48,6 +48,11 @@ public class BlackDuckResponsesTransformer {
     }
 
     public <T extends BlackDuckResponse> BlackDuckPageResponse<T> getResponses(PagedRequest pagedRequest, Class<T> clazz, boolean getAll) throws IntegrationException {
+        int maxPageCount = getAll ? Integer.MAX_VALUE : 1;
+        return getResponses(pagedRequest, clazz, maxPageCount);
+    }
+
+    public <T extends BlackDuckResponse> BlackDuckPageResponse<T> getResponses(PagedRequest pagedRequest, Class<T> clazz, int maxPageCount) throws IntegrationException {
         List<T> allResponses = new LinkedList<>();
         int totalCount = 0;
         int currentOffset = pagedRequest.getOffset();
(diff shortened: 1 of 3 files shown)


src/main/java/com/synopsys/integration/blackduck/service/BlackDuckResponsesTransformer.java
```diff
@@ -59,12 +64,14 @@ public BlackDuckResponsesTransformer(BlackDuckHttpClient blackDuckHttpClient, Bl
...
-            if (!getAll) {
+            if (maxPageCount <= 1) {
                 return new BlackDuckPageResponse<>(totalCount, allResponses);
             }
 
-            while (allResponses.size() < totalCount && currentOffset < totalCount) {
+            int currentPageCount = 1;
+            while (allResponses.size() < totalCount && currentOffset < totalCount && currentPageCount < maxPageCount) {
                 currentOffset += pagedRequest.getLimit();
+                currentPageCount += 1;
```
bamandel: I believe there is a shorter notation for this.

bamandel: currentPageCount++ is what I am referring to.

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckResponsesTransformer.java
```diff
@@ -59,12 +64,14 @@ public BlackDuckResponsesTransformer(BlackDuckHttpClient blackDuckHttpClient, Bl
             allResponses.addAll(blackDuckPageResponse.getItems());
 
             totalCount = blackDuckPageResponse.getTotalCount();
-            if (!getAll) {
+            if (maxPageCount <= 1) {
```
bamandel: This check doesn't seem necessary. Since you have the loop below, your loop should finish after its first iteration. Unless I'm missing something here.

JakeMathews: We always make a call for the first page first. We won't know the total number of items available to page until we make 1 request

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckResponsesTransformer.java
JakeMathews: You're right.

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckResponsesTransformer.java
JakeMathews: Perhaps we can make the loop smarter, but I didn't want to make too many changes to the behavior of this method.

PR #275, opened 2019-11-21
Bdio 2

JakeMathews:
Adds support for BDIO 2 to blackduck-common

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/service/BlackDuckService.java
+++ b/src/main/java/com/synopsys/integration/blackduck/service/BlackDuckService.java
@@ -58,7 +58,7 @@ public class BlackDuckService {
     public static final BlackDuckPath BOMIMPORT_PATH = new BlackDuckPath("/api/bom-import");
     public static final BlackDuckPath SCANSUMMARIES_PATH = new BlackDuckPath("/api/scan-summaries");
     public static final BlackDuckPath UPLOADS_PATH = new BlackDuckPath("/api/uploads");
-    public static final BlackDuckPath SCAN_DATA_PATH = new BlackDuckPath("/api/scan/data");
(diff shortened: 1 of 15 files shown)


JakeMathews:
Will be adding tests soon.

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckService.java
```diff
@@ -58,6 +58,7 @@
     public static final BlackDuckPath BOMIMPORT_PATH = new BlackDuckPath("/api/bom-import");
     public static final BlackDuckPath SCANSUMMARIES_PATH = new BlackDuckPath("/api/scan-summaries");
     public static final BlackDuckPath UPLOADS_PATH = new BlackDuckPath("/api/uploads");
+    public static final BlackDuckPath SCAN_DATA_PATH = new BlackDuckPath("/api/scan/data");
```
ketul-shukla: This api would need to have / at the end. 
eg: /api/scan/data/

ekerwin: I don't believe that is accurate, this is the correct way for the BlackDuckPath to be constructed.

JakeMathews: So I just tested and without the slash I get a 404 status code and with the slash I get a 201 status code. Not sure why this endpoint is different.

ekerwin: That sounds like a bug somewhere.

ekerwin:
I don't believe this code will be used for a Polaris upload, but if there are requirements to support that end point correctly, they should likely be spelled out in the doc for the api.

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckService.java
ketul-shukla: The API inside Blackduck is /scan/data/ that's the reason why I mentioned that.

PR #276, opened 2019-11-25
Ps media type support

psantos1113:
Changes to get the media type from the MediaTypeDirectory.

FYI we may want to rename MediaTypeDirectory to MediaTypeLookup since we are looking in a map for the media type.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/service/BlackDuckMediaTypeLookup.java
+++ b/src/main/java/com/synopsys/integration/blackduck/service/BlackDuckMediaTypeLookup.java
@@ -1,0 +1,17 @@
+package com.synopsys.integration.blackduck.service;
+
+import com.synopsys.integration.blackduck.api.core.BlackDuckResponse;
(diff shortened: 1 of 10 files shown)


src/main/java/com/synopsys/integration/blackduck/service/BlackDuckMediaTypeLookup.java
```diff
@@ -0,0 +1,17 @@
...
+public class BlackDuckMediaTypeLookup {
+    private final MediaTypeDiscovery mediaTypeDiscovery;
+
+    public BlackDuckMediaTypeLookup(MediaTypeDiscovery mediaTypeDiscovery) {
+        this.mediaTypeDiscovery = mediaTypeDiscovery;
+    }
+
+    public <T extends BlackDuckResponse> String findMediaType(Class<T> clazz) {
+        return mediaTypeDiscovery.determineMediaType(clazz).orElse(RequestFactory.DEFAULT_MEDIA_TYPE);
+    }
```
jamesrichard91: Wouldn't this make sense to be in MediaTypeDiscovery?

PR #280, opened 2020-02-04
Added boolean field licenseSearch to ScanCommand

crowleySynopsys:
Adding flag to ScanCommand that when set on Detect command line will pass scan CLI property onto the scanner


src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
```diff
@@ -219,6 +221,10 @@ public ScanCommand(final File installDirectory, final File outputDirectory, fina
                 }
             }
         }
+
+        if (licenseSearch) {
+            cmd.add("--license-search");
```
jamesrichard91: This option doesn't take any arguments?

crowleySynopsys: ya apparently it's just a boolean

PR #292, opened 2020-07-31
Ek media lookup

ekerwin:


src/main/java/com/synopsys/integration/blackduck/http/BlackDuckRequestBuilder.java
```diff
@@ -0,0 +1,158 @@
...
+import com.synopsys.integration.rest.HttpUrl;
+import com.synopsys.integration.rest.body.BodyContent;
+import com.synopsys.integration.rest.request.Request;
+
+import java.nio.charset.Charset;
+import java.util.Map;
+import java.util.Optional;
+import java.util.Set;
+
+import static com.synopsys.integration.blackduck.http.RequestFactory.*;
```
psantos1113, line 36: Should we be using wildcard imports?

PR #308, opened 2020-10-07
Ek add user agent

ekerwin:


src/main/java/com/synopsys/integration/blackduck/keystore/KeyStoreHelper.java
```diff
@@ -28,6 +50,10 @@ public KeyStoreHelper(final IntLogger logger) {
     public void updateKeyStoreWithServerCertificate(HttpUrl httpsServer, String keyStoreFilePath) {
         HttpsURLConnection httpsConnection = null;
         try {
+            if (!(httpsServer.url().openConnection() instanceof HttpsURLConnection)) {
```
jamesrichard91, line 53: Does this leave an open connection? Shouldn't this be closed?

PR #328, opened 2021-01-15
Jm project service predicate

JakeMathews:
Adds the ability for `ProjectService` to accept a `Predicate` for filtering results when getting all projects from Black Duck.


jamesrichard91 (approved):

gkillough (approved):

JakeMathews:
We should be using the more efficient `ProjectService::getAllMatches` method instead.

PR #339, opened 2021-03-02
Change project name to be unique

DMaxfield-BDS:


src/test/java/com/synopsys/integration/blackduck/comprehensive/recipe/Bdio2UploadRecipeTest.java
```diff
@@ -44,15 +44,13 @@
 @ExtendWith(TimingExtension.class)
 public class Bdio2UploadRecipeTest extends BasicRecipe {
     public static final String CODE_LOCATION_NAME = "bdio2 code location junit";
-    public static final NameVersion PROJECT = new NameVersion("blackduck-common", "test");
+    public static final NameVersion PROJECT = new NameVersion("bdio2uploadtest-" + System.currentTimeMillis(), "test");
```
jamesrichard91, line 47: If you really want this to be unique you could use UUID. And if this is always going to be a unique name, you don't need to check if the project exists on line 60

PR #359, opened 2021-07-29
BDIO 2 Project/Version Name

JakeMathews:
A project name and version is no longer required to publish bdio2 documents.
When supplied with a project and version name, headers will be added to the upload requests.

I have tested it with Detect, no changes are needed there.


stevebillings (approved):
Nice... I think the editor idea worked out really well.

PR #381, opened 2022-08-29
Sb retry bdio upload

stevebillings:
Retry IP bdio upload if it fails. (Earlier changes didn't account for the fact that an exception is thrown when there's a failure. With these changes it'll handle the exception case, or the no-exception case, in case integration-rest is changed in the future to not throw exceptions on failure.)


DanaMaxfield (approved):
I'm surprised no tests needed to be updated with this change. Can we add some for this?

stevebillings:
> I'm surprised no tests needed to be updated with this change. Can we add some for this?

Great point. Now done.

PR #387, opened 2022-10-31
add reduced persistence flags for signature scans

dterrysynopsys:
Adds flags to allow users to specify if unmatched files should be persisted or purged when running signature scans. This aligns with new features in BlackDuck 2022.10

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanBatch.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanBatch.java
@@ -16,6 +16,7 @@ import org.apache.commons.lang3.StringUtils;
 
 import com.synopsys.integration.blackduck.codelocation.signaturescanner.command.BlackDuckOnlineProperties;
 import com.synopsys.integration.blackduck.codelocation.signaturescanner.command.IndividualFileMatching;
+import com.synopsys.integration.blackduck.codelocation.signaturescanner.command.ReducedPersistence;
 import com.synopsys.integration.blackduck.codelocation.signaturescanner.command.ScanCommand;
 import com.synopsys.integration.blackduck.codelocation.signaturescanner.command.ScanPathsUtility;
 import com.synopsys.integration.blackduck.codelocation.signaturescanner.command.ScanTarget;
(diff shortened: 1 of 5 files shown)


src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
```diff
@@ -132,12 +134,25 @@ public List<String> createCommandForProcessBuilder(IntLogger logger, ScanPaths s
...
+	private void populateReducedPersistence(List<String> cmd) {
+		if (reducedPersistence != null) {
+        	if (reducedPersistence.equals(ReducedPersistence.DISCARD_UNMATCHED)) {
+        		cmd.add("--discard-unmatched-files");
+        	}
+        	if (reducedPersistence.equals(ReducedPersistence.RETAIN_UNMATCHED)) {
+        		cmd.add("--retain-unmatched-files");
+        	}
+        }
+	}
```
ChomickiM, line 154: May need to re-run the code formatter here to fix indentation

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanBatch.java
```diff
@@ -54,11 +55,12 @@ public static ScanBatchBuilder newBuilder() {
     private final boolean debug;
     private final boolean verbose;
     private final boolean isRapid;
+    private final ReducedPersistence reducedPersistence;
```
stevebillings, line 58: It took me a while to come up with a theory to explain why some sig scanner options go into BlackDuckOnlineProperties, and others go here in ScanBatch. I guess ScanBatch is for options that apply in offline mode as well as online (where-as BlackDuckOnlineProperties seems to be for online-only options).

Does reducedPersistence apply when Detect runs offline? If so, these changes seem fine to me.

stevebillings (approved):

stevebillings (changes requested):
I'm not actually requesting changes... I just have a question.

jppetrakis (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanBatch.java
dterrysynopsys: Interesting, thanks for the explanation. I think in this case the flag applies whether detect is run in online or offline mode. I used the --blackduck.offline.mode=true setting and ran the same scans I was doing to test online. I grabbed the BDIO and uploaded it and it seems like BD reacts in the appropriate way when I check the database. Going to leave the code as is based on this.

stevebillings (approved):
Cool, thanks.

ChomickiM (approved):

PR #391, opened 2022-12-01
add scanID information to BDIO responses

dterrysynopsys:
When we upload a BDIO BlackDuck will tell us the scanID it generated. Right now we drop it since we work in terms of code location names. This work adds the scanID to relevant output classes.

I'll leave it as a draft until I can get a released version of blackduck-common-api which is needed.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/upload/UploadOutput.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/upload/UploadOutput.java
@@ -17,26 +17,32 @@ import com.synopsys.integration.util.NameVersion;
 
 public class UploadOutput extends CodeLocationOutput {
     private final String response;
+    private final String scanId;
 
-    private UploadOutput(@Nullable NameVersion projectAndVersion, String codeLocationName, Result result, String response, String errorMessage, Exception exception) {
+    private UploadOutput(@Nullable NameVersion projectAndVersion, String codeLocationName, Result result, String response, String errorMessage, Exception exception, String scanId) {
         super(result, projectAndVersion, codeLocationName, 1, errorMessage, exception);
         this.response = response;
+        this.scanId = scanId;
     }
 
-    public static UploadOutput SUCCESS(@Nullable NameVersion projectAndVersion, String codeLocationName, String response) {
-        return new UploadOutput(projectAndVersion, codeLocationName, Result.SUCCESS, response, null, null);
+    public static UploadOutput SUCCESS(@Nullable NameVersion projectAndVersion, String codeLocationName, String response, String scanId) {
+        return new UploadOutput(projectAndVersion, codeLocationName, Result.SUCCESS, response, null, null, scanId);
     }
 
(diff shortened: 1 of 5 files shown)


jppetrakis (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/upload/UploadOutput.java
```diff
@@ -17,26 +17,32 @@
...
-    private UploadOutput(@Nullable NameVersion projectAndVersion, String codeLocationName, Result result, String response, String errorMessage, Exception exception) {
+    private UploadOutput(@Nullable NameVersion projectAndVersion, String codeLocationName, Result result, String response, String errorMessage, Exception exception, String scanId) {
         super(result, projectAndVersion, codeLocationName, 1, errorMessage, exception);
         this.response = response;
+        this.scanId = scanId;
     }
 
-    public static UploadOutput SUCCESS(@Nullable NameVersion projectAndVersion, String codeLocationName, String response) {
-        return new UploadOutput(projectAndVersion, codeLocationName, Result.SUCCESS, response, null, null);
+    public static UploadOutput SUCCESS(@Nullable NameVersion projectAndVersion, String codeLocationName, String response, String scanId) {
```
stevebillings, line 28: I think it would require some digging by someone writing new upload code to know that it's OK to pass null for ScanId. Seems better to have two versions of this SUCCESS method, one that with the scanId arg, one without. Or, at a minimum, I think the arg should be marked @Nullable.

src/main/java/com/synopsys/integration/blackduck/codelocation/upload/UploadOutput.java
```diff
@@ -17,26 +17,32 @@
...
     public static UploadOutput FAILURE(@Nullable NameVersion projectAndVersion, String codeLocationName, String response, String errorMessage, Exception exception) {
-        return new UploadOutput(projectAndVersion, codeLocationName, Result.FAILURE, response, errorMessage, exception);
+        return new UploadOutput(projectAndVersion, codeLocationName, Result.FAILURE, response, errorMessage, exception, null);
     }
 
     public Optional<String> getResponse() {
         return Optional.ofNullable(response);
     }
 
+	public String getScanId() {
```
stevebillings, line 44: I'd suggest @Nullable or Optional return value.

stevebillings (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/upload/UploadOutput.java
gopannabd: I too prefer overloading to handling nullable arguments.

gopannabd:
I do not agree to Black Duck generating the scan ID. The scan ID should be generated by the producer - here Detect.
This is a bad design that we are being forced to accommodate.

I will initiate a conversation on this in our next meeting. Under a separate ticket, we should generate the scan ID and all consumers should use that.

gopannabd:
I don't see an Approve button. Please consider this comment as approval.

DanaMaxfield (approved):

PR #392, opened 2022-12-05
Allow downstream clients to use the dependency scope field

romeara:
Add use of the dependency scope field when generating BDIO dependency representations

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,13 +20,14 @@ versionFile << version
 
 repositories {
     maven { url "https://sig-repo.synopsys.com/bds-bdio-release" }
+    mavenLocal()
 }
 
 dependencies {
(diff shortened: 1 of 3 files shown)


build.gradle
```diff
@@ -20,13 +20,14 @@ versionFile << version
 
 repositories {
     maven { url "https://sig-repo.synopsys.com/bds-bdio-release" }
+    mavenLocal()
```
romeara, line 23: Note: This PR is dependent/blocked by the release of integrations-bdio 26.0.8. Once that release exists, this PR can be updated to remove this mavenLocal line, and use a non-SNAPSHOT version. 
Given the lack of SNAPSHOT availability to the build pipeline, it is likely PR builds will fail for this request until this change is applied.

PR #401, opened 2023-05-26
version check for http headers

dterrysynopsys:
This work moves the version class files from the main detect code to this library. This allows us to have version checks at various levels of the code which more use cases are deeming necessary.

The version checks in this PR are then used to determine if we can safely omit the HTTP headers for project names and project versions as BD will pick up these values from the BDIO header instead. This fixes cases where non-ASCII characters were previously supplied in the HTTP headers causing REST calls to fail.

Note: this only conditionally adds REST headers for uploading intelligent persistent related BDIO files. It does not alter all areas where REST headers are used.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/bdio2/Bdio2FileUploadService.java
+++ b/src/main/java/com/synopsys/integration/blackduck/bdio2/Bdio2FileUploadService.java
@@ -95,8 +95,8 @@ public class Bdio2FileUploadService extends DataService {
     // project names and version names and do not need the REST headers which can have issues
     // with non-ASCII characters.
 	private boolean useOnlyBdioHeaders() {
-		if (blackDuckApiClient.getBlackDuckVersion() != null) {
-			BlackDuckVersion currentBlackDuckVersion = blackDuckApiClient.getBlackDuckVersion();
+		if (blackDuckApiClient.getBlackDuckVersion().isPresent()) {
+			BlackDuckVersion currentBlackDuckVersion = blackDuckApiClient.getBlackDuckVersion().get();
 			BlackDuckVersion requiresBdioHeadersVersion = new BlackDuckVersion(2023, 4, 1);
 			
 			return currentBlackDuckVersion.isAtLeast(requiresBdioHeadersVersion);
--- a/src/main/java/com/synopsys/integration/blackduck/bdio2/Bdio2FileUploadService.java
+++ b/src/main/java/com/synopsys/integration/blackduck/bdio2/Bdio2FileUploadService.java
@@ -20,6 +20,7 @@ import com.synopsys.integration.blackduck.exception.BlackDuckIntegrationExceptio
 import com.synopsys.integration.blackduck.service.BlackDuckApiClient;
 import com.synopsys.integration.blackduck.service.DataService;
 import com.synopsys.integration.blackduck.service.request.BlackDuckRequestBuilderEditor;
+import com.synopsys.integration.blackduck.version.BlackDuckVersion;
 import com.synopsys.integration.exception.IntegrationException;
 import com.synopsys.integration.log.IntLogger;
 import com.synopsys.integration.util.NameVersion;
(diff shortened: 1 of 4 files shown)


jahid1209 (approved):

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckApiClient.java
```diff
@@ -165,6 +168,14 @@ public HttpUrl executePostRequestAndRetrieveURL(BlackDuckResponseRequest request
             throw new IntegrationException(e.getMessage(), e);
         }
     }
+    
+    public BlackDuckVersion getBlackDuckVersion() {
+		return blackDuckVersion;
+	}
```
ChomickiM, line 174: This should be made to return an Optional<BlackDuckVersion> otherwise we risk a NPE here.

src/main/java/com/synopsys/integration/blackduck/bdio2/Bdio2FileUploadService.java
```diff
@@ -85,4 +91,18 @@ private Bdio2UploadResult uploadFiles(List<BdioFileContent> bdioFiles, @Nullable
         return jobExecutor.executeJob(bdio2UploadJob);
     }
 
+    // BlackDuck servers 2023.4.1 and later can use only BDIO header information for 
+    // project names and version names and do not need the REST headers which can have issues
+    // with non-ASCII characters.
+	private boolean useOnlyBdioHeaders() {
+		if (blackDuckApiClient.getBlackDuckVersion() != null) {
```
ChomickiM, line 98: Change the return of getBlackDuckVersion() to return an Optional<BlackDuckVersion> and validate isPresent() here instead.

src/main/java/com/synopsys/integration/blackduck/service/BlackDuckApiClient.java
```diff
@@ -39,13 +40,15 @@ public class BlackDuckApiClient {
     private final BlackDuckJsonTransformer blackDuckJsonTransformer;
     private final BlackDuckResponseTransformer blackDuckResponseTransformer;
     private final BlackDuckResponsesTransformer blackDuckResponsesTransformer;
+    private BlackDuckVersion blackDuckVersion;
 
-    public BlackDuckApiClient(BlackDuckHttpClient blackDuckHttpClient, BlackDuckJsonTransformer blackDuckJsonTransformer, BlackDuckResponseTransformer blackDuckResponseTransformer,
+	public BlackDuckApiClient(BlackDuckHttpClient blackDuckHttpClient, BlackDuckJsonTransformer blackDuckJsonTransformer, BlackDuckResponseTransformer blackDuckResponseTransformer,
```
ChomickiM, line 45: Slight formatting change, remove the extra tab

dterrysynopsys:
updated based on code review comments

gopannabd (approved):

src/main/java/com/synopsys/integration/blackduck/version/BlackDuckVersion.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public int getMinor() {
+        return minor;
+    }
+
+    public int getPatch() {
+        return patch;
+    }
+
+    public boolean isAtLeast(BlackDuckVersion other) {
```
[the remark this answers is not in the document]
dterrysynopsys: added explanation

src/main/java/com/synopsys/integration/blackduck/version/BlackDuckVersion.java
```diff
@@ -0,0 +1,49 @@
...
+        if (minor < other.getMinor()) {
+            return false;
+        }
+        if (patch > other.getPatch()) {
+            return true;
+        }
+        if (patch < other.getPatch()) {
+            return false;
+        }
+        return true;
```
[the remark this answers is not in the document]
dterrysynopsys: I like this idea but for now I'd prefer to leave the code as is given it has been holding up well.

PR #406, opened 2023-07-25
Guard against Windows API requests that exceed OS character limits

dterrysynopsys:
When making calls to the Scan CLI, it is possible for Detect to construct commands that exceed the Windows API's character length limitations. Instead of having the Scan CLI fail and Detect ungracefully handle the fallout, customers have asked to guard against the failure and log messages to describe the problem.


src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
```diff
@@ -58,6 +60,13 @@ public ScanCommandOutput call() {
             cmd.add(scanCommand.getTargetPath());
 
             commandToExecute = createPrintableCommand(cmd);
+            
+            if (OperatingSystemType.determineFromSystem().equals(OperatingSystemType.WINDOWS) 
+            		&& commandToExecute.length() > WINDOWS_CHARACTER_LIMIT) {
+        	    String errorMessage = "Unable to invoke the scan CLI as the length of the command would exceed the operating system limit.";
+        	    return ScanCommandOutput.FAILURE(projectAndVersion, codeLocationName, logger, scanCommand, commandToExecute, errorMessage, null);              
+            }
+
```
shantyk, line 69: Do we want to guard against other operating systems or is this only happening with Windows?

gopannabd: Only Windows.

src/test/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanCommandCallableTest.java
```diff
@@ -0,0 +1,65 @@
...
+        ScanBatchBuilder scanBatchBuilder = ScanBatch.newBuilder();
+        scanBatchBuilder.blackDuckUrl(new HttpUrl("http://fakeserver.com"));
+        scanBatchBuilder.blackDuckApiToken("fake_token");
+        scanBatchBuilder.addTarget(ScanTarget.createBasicTarget("fake_file_path"));
+        scanBatchBuilder.outputDirectory(tempDirectory.toFile());
+        ScanBatch scanBatch = scanBatchBuilder.build();
+        
+        // Create a very long command length so we are certain to blow past the Windows 32108 character API limit
+        IntEnvironmentVariables intEnvironmentVariables = Mockito.mock(IntEnvironmentVariables.class);
+        Mockito.when(intEnvironmentVariables.getValue("SCAN_CLI_OPTS")).thenReturn("x".repeat(38000));
```
shantyk, line 50: Can we use WINDOWS_CHARACTER_LIMIT here as well? I don't anticipate the Windows char limit will change in the future, but just to be consistent with the changes in ScanCommandCallable and not using a hard coded value here

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
dterrysynopsys: It's a good question. Windows runs into trouble around 32k. I believe mac and linux run into trouble at around 200k. Given that's pretty huge I didn't want to restrict things there and potentially run into trouble but I could certainly add a check if we think it worthwhile.

src/test/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanCommandCallableTest.java
```diff
@@ -0,0 +1,65 @@
...
+        // Mock to return we are working on Windows regardless of where the tests are run
+        Mockito.mockStatic(OperatingSystemType.class);
+        Mockito.when(OperatingSystemType.determineFromSystem()).thenReturn(OperatingSystemType.WINDOWS);
+        
+        List<ScanCommand> scanCommands = scanBatch.createScanCommands(tempDirectory.toFile(), mockedScanPathsUtility, intEnvironmentVariables);
+        ScanCommand scanCommand = scanCommands.get(0);
+		ScanCommandCallable scanCommandCallable = new ScanCommandCallable(logger, mockedScanPathsUtility, intEnvironmentVariables, scanCommand, false);
+		ScanCommandOutput output = scanCommandCallable.call();
+		
+		Assertions.assertTrue(output.getErrorMessage().get().equals("Unable to invoke the scan CLI as the length of the command would exceed the operating system limit."));
```
shantyk, line 61: Is the wording "scan cli" consistent with how we have other user visible msgs related to signature scans? I remember when I first got started with Detect, "scan cli" was something I kept hearing about but wasn't clear on after reading the public documentation. Just a minor comment, don't have an answer but thought I'd mention it!

shantyk (approved):
Looks good overall, minor comments/questions

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
shantyk: I'm fine with merging these changes and addressing issues with linux/Mac if and when that is reported

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
```diff
@@ -58,6 +60,13 @@ public ScanCommandOutput call() {
             cmd.add(scanCommand.getTargetPath());
 
             commandToExecute = createPrintableCommand(cmd);
+            
+            if (OperatingSystemType.determineFromSystem().equals(OperatingSystemType.WINDOWS) 
+            		&& commandToExecute.length() > WINDOWS_CHARACTER_LIMIT) {
```
andrian-synopsys, line 65: Here we are checking the length of commandToExecute yet it seems to me like it is not the actual command that ends up executed (its content is generated using `createPrintableCommand()` which changes it somewhat.

Based on the stack trace in the ticket a safer way to do this may be to try to execute the command and catch `java.io.IOException: CreateProcess error=206`. What do you think?

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
```diff
@@ -24,9 +24,11 @@
 import com.synopsys.integration.log.IntLogger;
 import com.synopsys.integration.util.IntEnvironmentVariables;
 import com.synopsys.integration.util.NameVersion;
+import com.synopsys.integration.util.OperatingSystemType;
 
 public class ScanCommandCallable implements Callable<ScanCommandOutput> {
     private static final List<String> DRY_RUN_FILES_TO_KEEP = Arrays.asList("data");
+    private static final int WINDOWS_CHARACTER_LIMIT = 32108;
```
andrian-synopsys: Mostly curious, how did you determine this number?

dterrysynopsys: Will answer in above reply.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
dterrysynopsys: Yes, cmd is the actual command that is run but it was a bit trickier to calculate the length of that than the also available commandToExecute. That one is changed as you say but I believe all that createPrintableCommand does is change sensitive password fields. For example, if the supplied password was xyz we would substitute *** for it...my reading of the code was that we would have the same number of characters either way but there is something to be said about calculating on cmd as that is the actual command and perhaps things would deviate in the future.

This raises the point the limit I'm using is a combination of research and observed behavior. I didn't find a nice boundary like exactly 32k characters. 32,764 often comes up when googling but I found I had to lower that a bit as I kept getting failures.

My main approach had been to go after the concept of check the command length and prevent approach mentioned in the Jira issue. I think that prevents a bit of wasteful invocation time but I'm kind of coming around to your idea as it seems a bit safer and cleaner. Thanks for raising this and starting this conversation! @gopannabd do you have thoughts here?

src/test/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanCommandCallableTest.java
sig-cpotts: Hi @dterrysynopsys I would lean towards either "Signature Scanner", which is the term used throughout the BD documentation, or you could also include the "Black Duck" portion, though that is likely inferred.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommandCallable.java
gopannabd: It is a good idea to capture the exception to ensure we have coverage across OSs.

dterrysynopsys: @gopannabd @andrian-synopsys could you please take another look? I switched to the catch approach. Some relevant information: a) this error appears to be Windows only but I didn't see much advantage to putting a windows check around this and b) I could have caught the IOException in its own catch but this made the error message construction a tiny bit easier but I could go either way. Thoughts?

gopannabd (approved):

andrian-synopsys (approved):

DanaMaxfield (approved):

PR #409, opened 2023-10-01
BoM reporting / blackduck-common

jenszo:
Much like the NOTICE and Vulnerability Report, we're in need to download the Bill of Material from Blackduck Hub without auxiliary scripting (python, bash) in our pipelines. However, this currently is not supported natively by synopsys-detect. 

In pursuit of integrating it into synopsys-detect eventually, we've set out to add the necessary parts to the following modules:
- blackduck-common-api
- blackduck-common
- blackduck-java-api-examples

Please consider this PR as one of three.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -23,7 +23,7 @@ repositories {
 }
 
 dependencies {
-    api 'com.synopsys.integration:blackduck-common-api:2023.4.2.2'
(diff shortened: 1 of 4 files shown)


gopannabd:
@jenszo Thank you for the suggested changes. The changes are being reviewed.

PR #416, opened 2024-02-13
Treat key-value pairs in scan command additional arguments as overrides

andrian-synopsys:
Prior to this change, all additional arguments to scan command were treated as addition with no regard to what arguments may already be set.

This change ensures that key-value pairs get overridden in order to avoid conflicts and ambiguity.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
@@ -174,8 +174,8 @@ public class ScanCommand {
         }
         List<Integer> indecesList = commandKeysToKeyRelatedIndices.get(key);
 
-        indecesList.add(command.size());
-        indecesList.add(command.size() + 1);
+        indecesList.add(command.size());     // track the index of the key
+        indecesList.add(command.size() + 1); // track the index of the value
 
         appendSingleArgument(key);
         appendSingleArgument(value);
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
@@ -1,7 +1,7 @@
 /*
  * blackduck-common
  *
- * Copyright (c) 2023 Synopsys, Inc.
+ * Copyright (c) 2024 Synopsys, Inc.
  *
  * Use subject to the terms and conditions of the Synopsys End User Software License and Maintenance Agreement. All rights reserved worldwide.
  */
@@ -11,6 +11,9 @@ import java.io.File;
 import java.util.ArrayList;
 import java.util.List;
 import java.util.Set;
+import java.util.HashSet;
+import java.util.Map;
+import java.util.HashMap;
 
 import org.apache.commons.lang3.StringUtils;
 import org.jetbrains.annotations.Nullable;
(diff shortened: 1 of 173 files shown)


devmehtasynopsys (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
```diff
@@ -87,188 +94,225 @@ public ScanCommand(File signatureScannerInstallDirectory, File outputDirectory,
...
             if (StringUtils.isNotBlank(argument)) {
-                cmd.add(argument);
+                if (commandKeysToKeyRelatedIndeces.containsKey(argument)) {
+                    overriddenArgIndexes.addAll(commandKeysToKeyRelatedIndeces.get(argument));
+                }
+                appendSingleArgument(argument);
+            }
+        }
+
+        if (overriddenArgIndexes.size() > 0) {
```
niravrsynopsys, line 209: Can be replaced with an `isEmpty` check

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
```diff
@@ -52,6 +55,10 @@ public class ScanCommand {
     private final String correlationId;
     private final String bomCompareMode;
 
+    private List<String> command;
+
+    private Map<String, List<Integer>> commandKeysToKeyRelatedIndeces;
```
niravrsynopsys, line 60: Possible typo here with `indeces`? Is that meant to be `indices`? I also see some variables names include `indexes` (example: overriddenArgIndexes) in this file so it might be easier for future maintenance to use the same term if it all means the same?

niravrsynopsys (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
```diff
@@ -87,188 +94,225 @@ public ScanCommand(File signatureScannerInstallDirectory, File outputDirectory,
...
+        if (!commandKeysToKeyRelatedIndices.containsKey(key)) {
+            commandKeysToKeyRelatedIndices.put(key, new ArrayList<>());
+        }
+        List<Integer> indecesList = commandKeysToKeyRelatedIndices.get(key);
 
-        return cmd;
+        indecesList.add(command.size());
+        indecesList.add(command.size() + 1);
+
+        appendSingleArgument(key);
```
gopannabd, line 180: Should this be instead appendKeyValuePair(...)?

andrian-synopsys: Here we're already in the `appendKeyValuePair()` method.
Or are you suggesting we have another one with a different set of arguments?

gopannabd (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScanCommand.java
gopannabd: I meant should this method or perhaps an additional method take a stream of key-value pairs? (...)

andrian-synopsys: I see. I am not convinced that it would make the code clearer.
When I was putting this together I was actually torn between doing what I did (calling `appendSingleArgument()`) or adding items to the list directly at these lines.
Do you think the latter would be better?

gopannabd: It is too early for this work to justify that level of complexity. We can revisit when we implement centralized configuration management.

PR #422, opened 2024-06-06
Update SignatureScannerZipInstaller to use tools download API (decoup…

shantyk:
This MR updates the endpoint we call on when downloading the latest scan-cli from BD. Depending on the BD server version, we will either call the new API (for version 2024.7.0 and newer) or continue to call the "old" API (versions 2024.4.1 and earlier). 
Example:
Old way of downloading scan-cli: 
**some-bd-instance.com/download/scan.cli-macosx.zip**
New way of downloading scan-cli: 
**some-bd-instance.com/api/tools/scan.cli.zip/versions/latest/platforms/macosx**

Docs for the "new" tools download API can be found at some-bd-instance.com/api-doc/public.html#_downloading_scan_tools (I recommend reviewing that before reviewing this MR). 

Changes in summary: 
- Call new api/tools API to download scan-cli ZIP if client is connected to BD server version 2024.7.0+ (version check is done by Detect before calling the code in this library, see: https://github.com/blackducksoftware/synopsys-detect/pull/1141). (The rest of the installer functionality remains unchanged)
- Handle 304s in cases where the scan-cli local installed version matches the version available from Black Duck. (Main difference between old and new APIs is that the version check happens at the endpoint rather than before we make the request for download)

Nothing changes about the way we handle running the scan once scan-cli is downloaded or if the tool was pointed to a local path, or if an install already exists (and is not outdated). (A summary of the log output changes can be found at the confluence page linked under IDETECT-4244)


src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ScannerZipInstaller.java
```diff
@@ -13,33 +13,38 @@
...
+    private static final String SCAN_CLI_TOOL_DOWNLOAD_URL = "api/tools/scan.cli.zip/versions/latest/";
 
     public static final String BLACK_DUCK_SIGNATURE_SCANNER_INSTALL_DIRECTORY = "Black_Duck_Scan_Installation";
-    public static final String VERSION_FILENAME = "blackDuckVersion.txt";
+    public static final String SCAN_CLI_VERSION_FILENAME = "scan-cli-version.txt";
 
     private final IntLogger logger;
-    private final SignatureScannerClient signatureScannerClient;
-    private final BlackDuckRegistrationService blackDuckRegistrationService;
+    private final SignatureScannerClient signatureScannerClient; // TOME maybe nuke?
```
shantyk, line 46: Testing why the download of the scan.cli.zip fails sometimes (at different places: sometimes during the copying from temporary zip to real directory, and sometimes at the following line that involves the SignatureScannerClient:                 `keyStoreHelper.updateKeyStoreWithServerCertificate(downloadUrl.url().getHost(), serverCertificate, scanPaths.getPathToCacerts());`)

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/NewApiScannerInstaller.java
```diff
@@ -0,0 +1,224 @@
...
+
+import com.synopsys.integration.blackduck.exception.BlackDuckIntegrationException;
+import com.synopsys.integration.exception.IntegrationException;
+import com.synopsys.integration.log.IntLogger;
+import com.synopsys.integration.rest.HttpUrl;
+import com.synopsys.integration.rest.response.Response;
+import com.synopsys.integration.util.CleanupZipExpander;
+import com.synopsys.integration.util.OperatingSystemType;
+
+public class NewApiScannerInstaller extends ApiScannerInstaller {
```
shantyk, line 32: Comparing this class to OldAPiScannerInstaller would be the easiest way to review as they are both meant to have the same exact functionality

src/test/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/NewApiScannerInstallerTest.java
```diff
@@ -0,0 +1,69 @@
...
+        assertTrue(scanPaths.isManagedByLibrary());
+        assertTrue(StringUtils.isNotBlank(scanPaths.getPathToJavaExecutable()));
+        assertTrue(StringUtils.isNotBlank(scanPaths.getPathToOneJar()));
+        assertTrue(StringUtils.isNotBlank(scanPaths.getPathToScanExecutable()));
+        assertTrue(new File(scanPaths.getPathToJavaExecutable()).canExecute());
+        assertTrue(new File(scanPaths.getPathToOneJar()).canExecute());
+        assertTrue(new File(scanPaths.getPathToScanExecutable()).canExecute());
+    }
+
+}
```
shantyk, line 69: The test instance we use is on a version that would not invoke the NewApiScannerInstaller. Troubleshooting that while looking into IDETECT-4402

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/NewApiScannerInstaller.java
andrian-sevastyanov: Since new tends to become old, I would suggest not using those words as part of class names.
Perhaps something like `VersionedApiScannerInstaller` and `NonVersionedApiScannerInstaller` could work?

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/OldApiScannerInstaller.java
```diff
@@ -93,18 +93,18 @@ public File installOrUpdateScanner() throws BlackDuckIntegrationException {
             }
         }
 
-        File scannerExpansionDirectory = new File(installDirectory, ScannerZipInstaller.BLACK_DUCK_SIGNATURE_SCANNER_INSTALL_DIRECTORY);
+        File scannerExpansionDirectory = new File(installDirectory, OldApiScannerInstaller.BLACK_DUCK_SIGNATURE_SCANNER_INSTALL_DIRECTORY);
```
andrian-sevastyanov, line 96: Looking at how some of these constants are used, it seems to me like it would be better to move some of them into the parent installer class (or some other central place). This would help avoid some of the duplication and also make it clearer that some of the properties of the two installers are identical.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/NewApiScannerInstaller.java
shantyk: That is a very good point, changed it to: 

NewApiScannerInstaller --> ToolsApiScannerInstaller (we call api/tools)
OldApiScannerInstaller --> ZipApiScannerInstaller (originally called the SignatureScannerZipInstaller... and the API path itself looks like /download/<operating system>.zip)

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/OldApiScannerInstaller.java
shantyk: Good point, thank you for raising this! Moved the BLACK_DUCK_SIGNATURE_SCANNER_INSTALL_DIRECTORY to be in the parent abstract class instead.

psantos1113 (approved):
There's a bit of duplicated code here that I think needs more time to improve. I'm not sure how exactly to best remove the duplicate code just yet. Since this is needed for a release of Detect I will work on some example changes in a separate branch and create a PR for it for the next major version of Blackduck common.  For now these changes appear to address the need.

andrian-sevastyanov (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -0,0 +1,223 @@
...
+import com.synopsys.integration.exception.IntegrationException;
+import com.synopsys.integration.log.IntLogger;
+import com.synopsys.integration.rest.HttpUrl;
+import com.synopsys.integration.rest.response.Response;
+import com.synopsys.integration.util.CleanupZipExpander;
+import com.synopsys.integration.util.OperatingSystemType;
+
+public class ToolsApiScannerInstaller extends ApiScannerInstaller {
+    // The tools API for downloading the scan-cli is called on by Detect for BD versions 2024.7.0 or newer
+    public static final BlackDuckVersion MIN_BLACK_DUCK_VERSION = new BlackDuckVersion(2024, 7, 0);
```
gopannabd, line 34: Value could be loaded from a .properties file.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -0,0 +1,223 @@
...
+        File scannerExpansionDirectory = new File(installDirectory, ToolsApiScannerInstaller.BLACK_DUCK_SIGNATURE_SCANNER_INSTALL_DIRECTORY);
+        scannerExpansionDirectory.mkdirs();
+
+        File versionFile = new File(scannerExpansionDirectory, ToolsApiScannerInstaller.SCAN_CLI_VERSION_FILENAME);
+        HttpUrl downloadUrl = getDownloadUrl();
+
+        try {
+            String scannerVersion;
+            if (!versionFile.exists()) {
+                logger.info("No version file exists, assuming this is new installation and the signature scanner should be downloaded.");
```
gopannabd, line 103: "Signature Scan CLI version file not detected. Default version will be downloaded."

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -0,0 +1,223 @@
...
+                scanExecutable.setExecutable(true);
+
+                logger.info("Black Duck Signature Scanner downloaded successfully.");
+                return latestScannerVersion;
+            } else if (response.getStatusCode() == 304) {
+                // If no need to update, response is HTTP 304 Not modified
+                logger.debug("Locally installed Signature Scanner version is up to date - skipping download.");
+                return localScannerVersion;
+            } else {
+                logger.debug("Unable to download Signature Scanner. Response code: " + response.getStatusCode() + " " + response.getStatusMessage());
```
gopannabd, line 204: Can we fall back to existing signature scanner if one exists?

PR #424, opened 2024-06-17
Manually increment version as the pipeline failed to do this

andrian-sevastyanov:
Manually increment version as the pipeline failed to do this.

As the next step and long-term fix the pipeline failure should be better understood and the pipeline should be fixed.


DanaMaxfield (changes requested):
Lines 3-6 need to be changed to be:

`apply from: 'https://raw.githubusercontent.com/blackducksoftware/integration-resources/master/gradle_common/buildscript-dependencies.gradle', to: buildscript`

gopannabd (approved):

DanaMaxfield (approved):

PR #425, opened 2024-07-03
Upgrade common-gradle-plugin to latest in order to stop uploading artifacts to Maven Central

andrian-sevastyanov:
Upgrade common-gradle-plugin to latest in order to stop uploading artifacts to Maven Central.

No evidence of recent blackduck-common usage based on Maven Central has been discovered. If a customer reaches out about being unable to download artifacts, we can point them to our Artifactory.


devmehtasynopsys (approved):

EULA.txt
andrian-sevastyanov, line 1: It is actually not clear to me why this file needs to be removed. I have done so based on @DanaMaxfield instructions in an email dated a couple of years ago. I wonder how this decision was made.

DanaMaxfield (changes requested):
Check with Legal on the EULA, but otherwise this looks good.

EULA.txt
andrian-sevastyanov: I chatted with @dmamidisynopsys and he confirmed that we have directive to remove these files from our repos.

DanaMaxfield (approved):

PR #431, opened 2024-07-25
Add csv archive support to signature scan invocation

dterrysynopsys:
This allows for passing a csv archive option to the signature scanner so that csv files can be created. Detect will begin doing this in offline mode so that users have access to the files.

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanBatch.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/ScanBatch.java
@@ -231,4 +231,8 @@ public class ScanBatch extends Stringable implements Buildable {
     public String getCorrelationId() {
         return correlationId;
     }
+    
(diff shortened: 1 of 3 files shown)


andrian-sevastyanov (approved):

ChomickiM (approved):

gopannabd:
Is --outputFormat customer facing?

dterrysynopsys:
> Is --outputFormat customer facing?

It is in the sense that you could run the scan CLI manually and specify it. This work is so that Detect can essentially ask for it to be added once we add a new specific Detect property.

gopannabd:
Will there be a Detect property that customer could set to enable this property? Or, is it going to be enabled by default in offline mode?

dterrysynopsys:
> Will there be a Detect property that customer could set to enable this property? Or, is it going to be enabled by default in offline mode?

Yes, this is the prep work for that. Another PR will be made on the Detect side to add the property.

PR #433, opened 2024-07-26
Update ToolsApiScannerInstaller to add BD server's certificate to the…

shantyk:
Update ToolsApiScannerInstaller to add the BD server's certificate to the Java keystore that comes inside the scan-cli ZIP. Otherwise we will get the following error when Detect invokes the downloaded scan-cli: 

2024-07-26 04:49:40 MDT ERROR [pool-3-thread-1-Stream Redirect Thread] --- ERROR: An error occurred - [ERR03_1001 Resource Allocation Error | Service: Unknown Service | Failed to establish the validity of the server's certificate (use --insecure to ignore).]  sun.security.provider.certpath.SunCertPathBuilderException: unable to find valid certification path to requested target

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
@@ -11,11 +11,13 @@ import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.nio.charset.Charset;
+import java.security.cert.Certificate;
 
 import com.synopsys.integration.blackduck.api.core.BlackDuckResponse;
 import com.synopsys.integration.blackduck.api.core.response.UrlSingleResponse;
 import com.synopsys.integration.blackduck.http.BlackDuckRequestBuilder;
 import com.synopsys.integration.blackduck.http.client.BlackDuckHttpClient;
+import com.synopsys.integration.blackduck.keystore.KeyStoreHelper;
 import com.synopsys.integration.blackduck.service.request.BlackDuckRequest;
 import com.synopsys.integration.blackduck.version.BlackDuckVersion;
 import org.apache.commons.compress.archivers.ArchiveException;
@@ -28,6 +30,8 @@ import com.synopsys.integration.rest.HttpUrl;
 import com.synopsys.integration.rest.response.Response;
 import com.synopsys.integration.util.CleanupZipExpander;
 import com.synopsys.integration.util.OperatingSystemType;
+
+import javax.net.ssl.HttpsURLConnection;
 
 public class ToolsApiScannerInstaller extends ApiScannerInstaller {
     // The tools API for downloading the scan-cli is called on by Detect for BD versions 2024.7.0 or newer
@@ -45,6 +49,7 @@ public class ToolsApiScannerInstaller extends ApiScannerInstaller {
     private final BlackDuckHttpClient blackDuckHttpClient;
     private final CleanupZipExpander cleanupZipExpander;
     private final ScanPathsUtility scanPathsUtility;
+    private final KeyStoreHelper keyStoreHelper;
     private final HttpUrl blackDuckServerUrl;
     private final OperatingSystemType operatingSystemType;
     private final File installDirectory;
@@ -54,6 +59,7 @@ public class ToolsApiScannerInstaller extends ApiScannerInstaller {
             BlackDuckHttpClient blackDuckHttpClient,
             CleanupZipExpander cleanupZipExpander,
             ScanPathsUtility scanPathsUtility,
+            KeyStoreHelper keyStoreHelper,
             HttpUrl blackDuckServerUrl,
             OperatingSystemType operatingSystemType,
             File installDirectory
@@ -66,6 +72,7 @@ public class ToolsApiScannerInstaller extends ApiScannerInstaller {
         this.blackDuckHttpClient = blackDuckHttpClient;
         this.cleanupZipExpander = cleanupZipExpander;
         this.scanPathsUtility = scanPathsUtility;
+        this.keyStoreHelper = keyStoreHelper;
         this.blackDuckServerUrl = blackDuckServerUrl;
         this.operatingSystemType = operatingSystemType;
         this.installDirectory = installDirectory;
@@ -194,6 +201,10 @@ public class ToolsApiScannerInstaller extends ApiScannerInstaller {
                 oneJar.setExecutable(true);
                 scanExecutable.setExecutable(true);
 
+                if (connectAndGetServerCertificate(downloadUrl)!= null) {
+                    keyStoreHelper.updateKeyStoreWithServerCertificate(downloadUrl.url().getHost(), connectAndGetServerCertificate(downloadUrl), scanPaths.getPathToCacerts());
+                }
+
                 logger.info("Black Duck Signature Scanner downloaded successfully.");
                 return latestScannerVersion;
             } else if (response.getStatusCode() == 304) {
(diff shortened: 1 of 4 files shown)


src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -220,4 +231,19 @@ private void removeOldVersionFileIfExists(File scannerExpansionDirectory) {
...
+            httpsConnection.connect();
+            Certificate[] certificates = httpsConnection.getServerCertificates();
+            httpsConnection.disconnect();
+
+            return certificates[0];
+        } catch (IOException e) {
+            logger.errorAndDebug("Could not get Black Duck server certificate which is required for managing the local keystore - communicating to the server will have to be configured manually: " + e.getMessage(), e);
+            return null;
+        }
+    }
```
shantyk, line 248: Connecting to the server to just fetch the certificate was done in the past: https://github.com/blackducksoftware/blackduck-common/pull/307/files 

Ideally using the connection we already establish for downloading the CLI would be ideal but looking at the DefaultBlackDuckHttpClient, I didn't see a clean way to do this.

src/test/java/com/synopsys/integration/blackduck/comprehensive/InstallAndRunSignatureScannerTestIT.java
```diff
@@ -73,7 +73,7 @@ public void cleanUp() {
         FileUtils.deleteQuietly(scannerDirectoryPath);
     }
 
-//    @Test
+    @Test
```
shantyk, line 76: Bringing back this test now that certificates are fixed on our butler instance and comprehensive pipeline is green again

shantyk, line 76: Bringing back this test now that certificates are fixed on our butler instance and comprehensive pipeline is green again

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -220,4 +231,19 @@ private void removeOldVersionFileIfExists(File scannerExpansionDirectory) {
...
+
+    private Certificate connectAndGetServerCertificate(HttpUrl httpsServer) {
+        HttpsURLConnection httpsConnection = null;
+        try {
+            httpsConnection = (HttpsURLConnection) httpsServer.url().openConnection();
+            httpsConnection.connect();
+            Certificate[] certificates = httpsConnection.getServerCertificates();
+            httpsConnection.disconnect();
+
+            return certificates[0];
```
andrian-sevastyanov, line 243: Any chance of this array being empty? If so, it might be good to check for that.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -194,6 +201,10 @@ protected String downloadSignatureScanner(File scannerExpansionDirectory, HttpUr
                 oneJar.setExecutable(true);
                 scanExecutable.setExecutable(true);
 
+                if (connectAndGetServerCertificate(downloadUrl)!= null) {
```
andrian-sevastyanov, line 204: It would be good to not call this redundantly: store the return value in a variable and use it for both the if condition and the update call.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -194,6 +201,10 @@ protected String downloadSignatureScanner(File scannerExpansionDirectory, HttpUr
...
+                    keyStoreHelper.updateKeyStoreWithServerCertificate(downloadUrl.url().getHost(), connectAndGetServerCertificate(downloadUrl), scanPaths.getPathToCacerts());
```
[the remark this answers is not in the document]
shantyk: Looking at the way the old API works, it does the same thing. Please see ZipApiScannerInstaller, particularly the following lines: 

```
Certificate serverCertificate = signatureScannerClient.getServerCertificate();

keyStoreHelper.updateKeyStoreWithServerCertificate(downloadUrl.url().getHost(), serverCertificate, scanPaths.getPathToCacerts());
```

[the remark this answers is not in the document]
andrian-sevastyanov: Yeah, I find that concerning as well.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
shantyk: I am not clear on how to do that since the download call (to _update_ the scan-cli) does not explicitly establish a connection here for me to hold on to (I'm not entirely clear on its lifetime either, though I believe it is until the end of the try-with-resources block).

andrian-sevastyanov: What I had in mind is:


```
Certificate cert = connectAndGetServerCertificate(downloadUrl);

if (cert != null) {
    keyStoreHelper.updateKeyStoreWithServerCertificate(downloadUrl.url().getHost(), cert, scanPaths.getPathToCacerts());
}
```

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
dterrysynopsys: Could we maybe have Chris review this error message? If this fails it might be better to try the download again. It would be pretty difficult to configure manually?

dterrysynopsys (approved):
Approving based on discussions. Please make Andrian's suggested optimization to avoid two connections.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
shantyk: When I first saw this I thought "oh? that's an option? how would one do that?" ... Not sure if it's mentioned in docs somewhere. 

@sig-cpotts Thoughts?   


This wording is a slight variation on what is logged today for essentially the same error through the "original" flow/API: https://github.com/blackducksoftware/blackduck-common/blob/84f9db2dec10892841e243c9db4fdfab7514c72b/src/main/java/com/synopsys/integration/blackduck/keystore/KeyStoreHelper.java#L45

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
shantyk: Oh I misunderstood your original comment. Thanks for noticing this, done!

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
shantyk: I went with:
```
try {
...
if (certificates.length > 0) {
                return certificates[0];
            } else {
                throw new IOException();
            }
        } catch (IOException e) {
            logger.errorAndDebug("Could not get Black Duck server certificate which is required for managing the local keystore - communicating to the server will have to be configured manually: " + e.getMessage(), e);
            return null;
        } 
```

andrian-sevastyanov (approved):

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
sig-cpotts: "How would one configure that manually" is for sure a valid question a user may have. : )
Would be nice to indicate in some brief way what that means, i.e. manually add server cert to JRE instance keystore. (Though there is a limit to how much detail can be in an error message.)
If there is potential that downloading again might resolve the issue, we should definitely indicate that to help the user save time/effort.
(We have a somewhat related note in the "Common Problems" section of the doc: https://sig-product-docs.synopsys.com/bundle/integrations-detect/page/troubleshooting/solutions.html#ariaid-title17)

PR #434, opened 2024-07-29
Using a single version file regardless of API used for download.

shantyk:
For simplicity sake, just using one text file: blackDuckVersion.txt regardless of which API was used to download the scanner. 

This is the same version file as we have always used, so as far as the Detect output files are concerned, all is the same regardless of which API was used under the hood .
  
IDETECT-4434

### DIFF
--- a/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
+++ b/src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
@@ -43,8 +43,6 @@ public class ToolsApiScannerInstaller extends ApiScannerInstaller {
     private static final String LINUX_PLATFORM_PARAMETER_VALUE = "linux";
     private static final String WINDOWS_PLATFORM_PARAMETER_VALUE = "windows";
 
-    public static final String SCAN_CLI_VERSION_FILENAME = "scan-cli-version.txt";
-
     private final IntLogger logger;
     private final BlackDuckHttpClient blackDuckHttpClient;
     private final CleanupZipExpander cleanupZipExpander;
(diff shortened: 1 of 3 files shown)


src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -220,20 +217,6 @@ protected String downloadSignatureScanner(File scannerExpansionDirectory, HttpUr
...
-    private void removeOldVersionFileIfExists(File scannerExpansionDirectory) {
-        try {
-            File oldVersionFile = new File(scannerExpansionDirectory, ZipApiScannerInstaller.VERSION_FILENAME);
-            oldVersionFile.delete();
-        } catch (Exception e) {
-            logger.trace("Could not delete old Signature Scanner version file.");
-        }
-    }
-
-
```
shantyk, line 236: Instead of adding yet another "deleteNewVersionFileIfExists()" method ... we just go back to using blackDuckVersion.txt (old version filename) and not worry about which API we used. We update the same version file with each download.

src/main/java/com/synopsys/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -220,20 +217,6 @@ protected String downloadSignatureScanner(File scannerExpansionDirectory, HttpUr
         }
     }
 
-    /**
-     * Deletes legacy version file that may be left behind if client had previously installed scan-cli using {@link ZipApiScannerInstaller}
```
dterrysynopsys, line 224: not sure if this is really a concern, but could we potentially be leaving this file behind before all versions of the CLI move to this new file approach?

shantyk: So with this MR, we no longer have "old version file" vs "new version file".  We continue to use the original file, "blackDuckVersion.txt" as we always have. Meaning: 

Regardless of which API ended up being called, whether through ToolsApiScannerInstaller or ZipApiScannerInstaller, both execution paths in the code will update this version file. Before this MR, I had gone with an approach where I create and maintain a new "scan-cli-version.txt" file for when the new API is being called. This created problems because you had to see "ok am I on the execution path for the new API or the old API? Which file should I read and which file should I make sure to delete?" which complicated the code and Mayank found a bug as well for a use case I hadn't thought of. To fix that bug, after discussing at standup, using just one file would be best

gopannabd (approved):

dterrysynopsys (approved):

niravrsynopsys (approved):

PR #442, opened 2024-08-22
Changes related to refactoring and fixing ITs.

shantyk:
Fixes an intermittently failing IT: IDETECT-4473

Please see ticket comment for details but in summary: Project/project version creations followed immediately by a deletion leads to a race condition. Other notification types appear to come from Hub backend and I did not see any APIs this test case could call on to trigger one of the other available types of notifications. The DELETE has been moved to the end of the test case as a clean up. Only one version is created instead of two because even after notifications are delivered and asserted, a race condition still happened when cleaning up.

### DIFF
--- a/src/test/java/com/synopsys/integration/blackduck/comprehensive/NotificationsTestIT.java
+++ b/src/test/java/com/synopsys/integration/blackduck/comprehensive/NotificationsTestIT.java
@@ -2,7 +2,6 @@ package com.synopsys.integration.blackduck.comprehensive;
 
 import static org.junit.jupiter.api.Assertions.assertEquals;
 
-import java.io.File;
 import java.time.temporal.ChronoUnit;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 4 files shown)


src/test/java/com/synopsys/integration/blackduck/comprehensive/NotificationsTestIT.java
```diff
@@ -68,8 +75,23 @@ public void testProjectNotifications() throws IntegrationException, InterruptedE
...
+         *     BOM_EDIT,
+         *     LICENSE_LIMIT,
+         *     POLICY_OVERRIDE,
+         *     PROJECT,
+         *     PROJECT_VERSION,
+         *     RULE_VIOLATION,
+         *     RULE_VIOLATION_CLEARED,
+         *     VERSION_BOM_CODE_LOCATION_BOM_COMPUTED,
+         *     VULNERABILITY;
+         */
```
shantyk, line 94: @DanaMaxfield I was going to try the 3rd TODO comment and maybe see if I can get the pipeline green again? Seems more of a solid solution than option 1

src/test/java/com/synopsys/integration/blackduck/comprehensive/NotificationsTestIT.java
```diff
@@ -100,6 +96,10 @@ public void testProjectNotifications() throws IntegrationException, InterruptedE
         }
 
         assertEquals(expectedKeys, foundKeys);
+
+        // CLEAN UP
+        blackDuckApiClient.delete(projectVersionWrapper2.getProjectVersionView());
+        blackDuckApiClient.delete(projectVersionWrapper.getProjectView());
```
shantyk, line 102: @DanaMaxfield We had agreed on replacing the deletes with an update notification, but after lots of trial I could not get that to happen (explanation below). So I've decided to just have this test confirm two CREATE notifications are created. Then at the very end, we do the delete (by which point we would hope the soft created project versions have been updated in the database already). 


I don't know if an "UPDATE" notification is ever created for any reason. Though ProjectNotificationContent has three types of possible OperationTypes: 
CREATE
DELETE
UPDATE

Doing an update to a project did not yield an "... UPDATE ....' notification. I also did not find a notification like that when looking at examples on a couple of test instances: 

Screenshot 2024-09-05 at 3 23 45 PM

For example I tried updating different project fields (any that did not require lots more state set up like creating user groups, policies, etc). I would get a log entry like saying an update was performed, and indeed I would see the change in the UI but not "UPDATE" notification would be created.

psantos1113: I don't believe an update notification to a project is possible.  A project is created or deleted and updates come from BomEditNotifications which aren't displayed in the UI.  They are a notification type that Alert uses and you need to filter for that notification type when using the REST API calls.

psantos1113: Operation types are the types of REST API calls you can make against the resource.  You can create a project notification, update a project notification i.e. mark it as viewed, and delete the notification.

DanaMaxfield (approved):

PR #448, opened 2024-10-14
add online csv mode

dterrysynopsys:
Add csv option conditionally based on if we are in online or offline mode.


psantos1113 (approved):

gopannabd (approved):

gopannabd:
Does this have a reference JIRA? If not, best to create one for reference.

dterrysynopsys:
> Does this have a reference JIRA or is it part of IDETECT-4373?

I've been considering it part of 4373. It's required by the Detect core changes which I'll be making a PR for soon.

PR #450, opened 2025-01-16
ComponentService: Fix-up obtaining a component view for a search result

fviernau:
The request construction accidentally uses the path to the `version` view, instead of to the `component` view. So, in case the given search result matches a `component`, but not a `version`, the result is empty. Fix that by using the right variable.

### DIFF
--- a/src/main/java/com/blackduck/integration/blackduck/service/dataservice/ComponentService.java
+++ b/src/main/java/com/blackduck/integration/blackduck/service/dataservice/ComponentService.java
@@ -91,8 +91,8 @@ public class ComponentService extends DataService {
     }
 
     public Optional<ComponentView> getComponentView(ComponentsView searchResult) throws IntegrationException {
-        if (StringUtils.isNotBlank(searchResult.getVersion())) {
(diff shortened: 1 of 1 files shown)


fviernau:
@shantyk can you make a review?

shantyk:
> @shantyk can you make a review? Or @dterrybd ?

Hi @fviernau, thank you for raising this and I apologize for the long delay (I had made note to review when this was first opened but it fell through the cracks). @dterrybd and I have looked into this and confirmed with other consumers of this library that the proposed fix is good to go. Thanks for the find and taking the time to contribute!

shantyk (approved):

fviernau:
> @dterrybd and I have looked into this and confirmed with other consumers of this

Thank you! (Maybe while it it, you can have a look at my other open PR for `integration-bdio` too? ;-)

dterrybd:
@fviernau This has been merged and released. You can find it in version 67.0.10 of the library. Thank you, we'll try to look at the integration-bdio PR in the near future.

PR #452, opened 2025-02-20
Bump bd-common-api version to 2023.4.2.8

shantyk:
Pending release pipeline, brings in https://github.com/blackducksoftware/blackduck-common-api/pull/64

Due to IDETECT-4497, we are using release/v2023.4.2.7.x in bd-common-api as the base branch for IDETECT-4262.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,7 +34,7 @@ repositories {
 }
 
 dependencies {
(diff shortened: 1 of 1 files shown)


dterrybd (approved):
I think this is relatively safe, since it doesn't look like a generation was actually done, just seems like a few rounds of bumping libraries...so I don't think there is much risk here.

PR #453, opened 2025-03-13
Reduce excessive logs when non-admin users fetch registration ID

zahidblackduck:
**JIRA Ticket:**
IDETECT-4209

**Description:**
This merge request addresses the issue of excessive warning logs generated when non-admin users attempt to fetch the registration ID on the endpoint `/api/registration`. 
~The solution involves using a `ThreadLocal` variable to manage the state of whether the `getRegistrationId()` method has been called before or not by a non-syadmin user. This ensures that repeated attempts to fetch the registration ID are avoided, thereby significantly reducing unnecessary warning logs.~

**Update 1:**
`BlackDuckRegistrationService` has been refactored. Now, it follows an eager approach. `getBlackDuckServerData()` method has been overloaded to accept `isAdmin` boolean flag. ~The passed boolean `isAdmin` flag is stored in a static local variable.
Before calling `getRegistrationId()`, a check is done whether `isAdmin` is false and skip the call accordingly.~
This avoids unnecessary calls to the method `getRegistrationId()`, thus reducing excessive logs in hub webapp container due to call to the endpoint `/api/registration`.

**Update 2:**
In this update, the `getBlackDuckServerData()` method has been overloaded to accept a boolean flag `isAdminOperationAllowed`. This flag controls access to the `getRegistrationId()` method, which triggers a call to the `/api/registration` endpoint. To restrict this call, the flag should be set accordingly.

Detect can now pass this flag when invoking the method, as it determines the user's admin status during connectivity checks, retains it for the duration of the scan, and passes it during phone-home operations.

The original `getBlackDuckServerData()` method (with no arguments) has been retained for backward compatibility.

**_This fix is required for resolving the linked Detect issue._**

### DIFF
--- a/src/main/java/com/blackduck/integration/blackduck/phonehome/BlackDuckPhoneHomeHelper.java
+++ b/src/main/java/com/blackduck/integration/blackduck/phonehome/BlackDuckPhoneHomeHelper.java
@@ -69,8 +69,12 @@ public class BlackDuckPhoneHomeHelper {
     }
 
     public PhoneHomeResponse handlePhoneHome(String integrationRepoName, String integrationVersion, Map<String, String> metaData, String... artifactModules) {
+        return handlePhoneHome(integrationRepoName, integrationVersion, metaData, true, artifactModules);
+    }
+
+    public PhoneHomeResponse handlePhoneHome(String integrationRepoName, String integrationVersion, Map<String, String> metaData, boolean isAdminOperationAllowed, String... artifactModules) {
         try {
-            PhoneHomeRequestBody phoneHomeRequestBody = createPhoneHomeRequestBody(integrationRepoName, integrationVersion, metaData, artifactModules);
+            PhoneHomeRequestBody phoneHomeRequestBody = createPhoneHomeRequestBody(integrationRepoName, integrationVersion, metaData, isAdminOperationAllowed, artifactModules);
             return phoneHomeService.phoneHome(phoneHomeRequestBody, getEnvironmentVariables());
         } catch (Exception e) {
             logger.debug("Problem phoning home: " + e.getMessage(), e);
@@ -79,13 +83,15 @@ public class BlackDuckPhoneHomeHelper {
     }
 
     private PhoneHomeRequestBody createPhoneHomeRequestBody(String integrationRepoName, String integrationVersion, Map<String, String> metaData, String... artifactModules) {
+        return createPhoneHomeRequestBody(integrationRepoName, integrationVersion, metaData, true, artifactModules);
+    }
+
+    private PhoneHomeRequestBody createPhoneHomeRequestBody(String integrationRepoName, String integrationVersion, Map<String, String> metaData, boolean isAdminOperationAllowed, String... artifactModules) {
         String registrationKey = PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE;
         String blackDuckUrl = PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE;
         String blackDuckVersion = PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE;
 
         try {
-            boolean isAdminOperationAllowed = Boolean.parseBoolean(metaData.getOrDefault("isAdminOperationAllowed", "false"));
-            metaData.remove("isAdminOperationAllowed");
             BlackDuckServerData blackDuckServerData = blackDuckRegistrationService.getBlackDuckServerData(isAdminOperationAllowed);
             registrationKey = blackDuckServerData.getRegistrationKey().orElse(PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE);
             blackDuckUrl = blackDuckServerData.getUrl().string();
@@ -111,5 +117,4 @@ public class BlackDuckPhoneHomeHelper {
         }
         return Collections.emptyMap();
     }
-
 }
--- a/src/main/java/com/blackduck/integration/blackduck/phonehome/BlackDuckPhoneHomeHelper.java
+++ b/src/main/java/com/blackduck/integration/blackduck/phonehome/BlackDuckPhoneHomeHelper.java
@@ -84,7 +84,8 @@ public class BlackDuckPhoneHomeHelper {
         String blackDuckVersion = PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE;
 
         try {
-            BlackDuckServerData blackDuckServerData = blackDuckRegistrationService.getBlackDuckServerData();
+            boolean isAdminOperationAllowed = Boolean.parseBoolean(metaData.getOrDefault("isAdminOperationAllowed", "false"));
+            BlackDuckServerData blackDuckServerData = blackDuckRegistrationService.getBlackDuckServerData(isAdminOperationAllowed);
             registrationKey = blackDuckServerData.getRegistrationKey().orElse(PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE);
             blackDuckUrl = blackDuckServerData.getUrl().string();
             blackDuckVersion = blackDuckServerData.getVersion();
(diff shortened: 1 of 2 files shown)


src/main/java/com/blackduck/integration/blackduck/service/dataservice/BlackDuckRegistrationService.java
```diff
@@ -45,11 +46,23 @@ public BlackDuckServerData getBlackDuckServerData() throws IntegrationException
...
         String registrationId = null;
         try {
-            // We need to wrap this because this will most likely fail unless they are running as an admin
-            registrationId = getRegistrationId();
+            if (isRegistrationIdFetchAllowed()) {
+                // We need to wrap this because this will most likely fail unless they are running as an admin
+                registrationId = getRegistrationId();
+            }
         } catch (IntegrationException e) {
+            // The user is not admin. So, this should prevent the fetching of the registrationId again
```
andrian-sevastyanov, line 54: I wonder if there are other situations in which `IntegrationException` could be encountered here, not only permissions issues. Is there perhaps something within the exception that can be checked that that's indeed the case?

shantyk (approved):

src/main/java/com/blackduck/integration/blackduck/service/dataservice/BlackDuckRegistrationService.java
zahidblackduck: I see. This exception could be thrown for any other reason too. I was thinking whether we could use any utility method like `isAdmin(...)` that would return a boolean and based on that we could evaluate this state setting/updating instead of setting that inside the catch.
The state could also be passed from the `detect` to the `blackduck-common` cause detect has the knowledge of the user who is trying to call this method via `BlackDuckRegistrationService#getBlackDuckServerData()` method.

shantyk: I believe if we get a 403 FORBIDDEN, setting isRegistrationIdFetchAllowed to false and not trying to call that endpoint again makes sense since the user permissions are unlikely to change during the rest of the scan run. But for different responses/exceptions, maybe we keep the behaviour the same?

zahidblackduck: > I wonder if there are other situations in which `IntegrationException` could be encountered here, not only permissions issues. Is there perhaps something within the exception that can be checked to confirm that that's indeed the underlying reason for the exception?

I've updated the code a bit here. Setting the `isRegistrationIdFetchAllowed` based on the `blackDuckErrorCode`. If the exception contains `core.rest.unauthorized` code, we set this variable `false`.

src/main/java/com/blackduck/integration/blackduck/service/dataservice/BlackDuckRegistrationService.java
```diff
@@ -24,6 +24,7 @@ public class BlackDuckRegistrationService extends DataService {
     private final UrlSingleResponse<RegistrationView> registrationResponse = apiDiscovery.metaRegistrationLink();
     private final UrlSingleResponse<CurrentVersionView> currentVersionResponse = apiDiscovery.metaCurrentVersionLink();
     private final HttpUrl blackDuckUrl;
+    private static final ThreadLocal<Boolean> isSystemAdministrator = ThreadLocal.withInitial(() -> true);
```
andrian-sevastyanov, line 27: Setting this to `true` initially could be misleading to someone looking at this code down the road and that could lead to unintended consequences.
Renaming the variable to something like `attemptAdminOperations` might be safer and more correct.

zahidblackduck: Thanks, Andrian, for the nice suggestion. I've renamed the variable to `adminOperationAttempted` and refactored the `getBlackDuckServerData(boolean)` method to handle most of the logic, with `getBlackDuckServerData()` now delegating to it.  

However, I believe we still need a variable to track the state of the user. This method is invoked multiple times throughout the application's execution while the connection remains active. If we were to pass the `isAdmin` flag explicitly for an user each time this method is called, we'd need to determine the user's admin status beforehand.

In detect, we can retrieve this information in `BlackDuckConnectivityChecker` based on the user's role. However, in `BlackDuckPhoneHomeHelper`, where `getBlackDuckServerData()` is also called during phone homing, user details are not directly available. To fetch them, we'd need access to `UserService`, which isn't currently available there.  

Given this, passing the `isAdmin` flag when determining connectivity in `BlackDuckConnectivityChecker#determineConnectivity(..)` and maintaining that state throughout the runtime of Detect seems like a better approach as of now.

src/main/java/com/blackduck/integration/blackduck/service/dataservice/BlackDuckRegistrationService.java
```diff
@@ -24,6 +24,7 @@ public class BlackDuckRegistrationService extends DataService {
     private final UrlSingleResponse<RegistrationView> registrationResponse = apiDiscovery.metaRegistrationLink();
     private final UrlSingleResponse<CurrentVersionView> currentVersionResponse = apiDiscovery.metaCurrentVersionLink();
     private final HttpUrl blackDuckUrl;
+    private static final ThreadLocal<Boolean> adminOperationAttempted = ThreadLocal.withInitial(() -> false);
```
andrian-sevastyanov, line 27: Looks like you renamed the variable AND flipped the initial value.
This means that we will not be preserving the original functionality of `getBlackDuckServerData()`.
As far as functionality goes, we should probably keep it the same as it was since we don't know all the usages.

zahidblackduck: I see. Let me re-evaluate this. Thanks for pointing out the concern.

zahidblackduck: Hi @andrian-sevastyanov, it seemed logical to me to set the initial value of the variable to `false` in the latest changes.
Cause, if the user is not `sysadmin`, why allow the user to make call to the endpoint `/api/registration` to fetch the registration id at all? If the user is not sysadmin, the `getBlackDuckServerData()` method should be functional as it was before.
For reference, there's a comment just above the codeblock where registration id is fetched,
```java
    // We need to wrap this because this will most likely fail unless they are running as an admin
    registrationId = getRegistrationId();
```

andrian-sevastyanov: A hypothetical scenario I'm thinking about is:
- non-Detect app that uses this code that we don't control (maybe Alert?)
- the app calls `getBlackDuckServerData()` while in admin mode
- the app maintainers are not aware of the change we are making and so they do not pass in the needed `isAdmin` value

This is why we should probably preserve the original functionality of `getBlackDuckServerData()` when no flag has been passed in.

dterrybd (approved):

andrian-sevastyanov (approved):

src/main/java/com/blackduck/integration/blackduck/phonehome/BlackDuckPhoneHomeHelper.java
```diff
@@ -84,7 +84,9 @@ private PhoneHomeRequestBody createPhoneHomeRequestBody(String integrationRepoNa
         String blackDuckVersion = PhoneHomeRequestBody.UNKNOWN_FIELD_VALUE;
 
         try {
-            BlackDuckServerData blackDuckServerData = blackDuckRegistrationService.getBlackDuckServerData();
+            boolean isAdminOperationAllowed = Boolean.parseBoolean(metaData.getOrDefault("isAdminOperationAllowed", "false"));
```
andrian-sevastyanov, line 87: Have you considered changing this class more similarly to how `BlackDuckRegistrationService` was modified?

In other words: add overloaded versions of public methods and pass in boolean values through there.

andrian-sevastyanov, line 87: If this is the approach we're taking, it would be good to have a constant defined for `isAdminOperationAllowed`. That way it's less hidden and could be referenced from other classes.

zahidblackduck: I see. That way we can keep the original implementation intact. I'll try that out.

psantos1113 (approved):

ChomickiM (approved):

DanaMaxfield (approved):

PR #462, opened 2025-06-25
Fix Potential bug

devmehtabd:
This PR fixes potential which can be caused if the blackduck url is not passed with a `/` at the end.


andrian-sevastyanov (approved):

andrian-sevastyanov (approved):

src/main/java/com/blackduck/integration/blackduck/bdio2/Bdio2UploadJob.java
```diff
@@ -75,6 +75,9 @@ public void attemptJob() throws IntegrationException {
                 uploadUrl = new HttpUrl(headerResponse.getHeaderValue("location"));
                 scanId = parseScanIdFromUploadUrl(uploadUrl.string());
             } else {
+                if(!blackDuckUrl.string().endsWith("/")) {
```
DanaMaxfield, line 78: Since blackDuckUrl is nullable, do you need to perform a not null check in this block?

PR #464, opened 2025-07-28
Check architecture and OS before downloading scan-cli

devmehtabd:
Added check for determining os architecture and type of OS before downloading scan-cli to support new endpoints for downloading scan cli as mentioned in IDETECT-4747 and IDETECT-4753.


src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -140,8 +144,14 @@ protected HttpUrl getDownloadUrl() throws BlackDuckIntegrationException {
             platform = MAC_PLATFORM_PARAMETER_VALUE;
         } else if (OperatingSystemType.WINDOWS == operatingSystemType) {
             platform = WINDOWS_PLATFORM_PARAMETER_VALUE;
-        } else {
+        } else if (OperatingSystemType.LINUX == operatingSystemType && !isAlpineLinux()) {
```
dterrybd, line 147: This is likely personal preference but I kind of like checking if it IS Alpine Linux and specifically setting the platform to Alpine if so. Then the else could be the fallback where you set the platform to general linux if not. Reads a bit clearer to me but I'm really fine either way.

andrian-sevastyanov (approved):

PR #468, opened 2025-08-14
Add scanCliArchitecture file for tracking architecture downloaded during previous runs

devmehtabd:
This PR adds a new architecture file along with version file in the same location where scan cli is stored. This will handle the case when arm64 users will try to upgrade from 10.6.0 to 10.7.0 version of Detect and have locally cached scan cli. Currently there is no way for Detect to differentiate the architecture between old and new scan cli for us to be able to update the locally stored scan cli.

There are two new conditions when fresh download will get triggered, if the architecture file does not exist and if it exists the current osArchitecture value is not equal to the one stored in file. Please let me know if there are some conditions which I am missing.

### DIFF
--- a/src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
+++ b/src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
@@ -133,7 +133,7 @@ public class ToolsApiScannerInstaller extends ApiScannerInstaller {
             // Initiate a fresh download and update architecture value in file
             if(!localArchitecture.equals(osArchitecture)) {
                 localScannerVersion = "";
-                FileUtils.writeStringToFile(versionFile, osArchitecture, Charset.defaultCharset());
+                FileUtils.writeStringToFile(architectureFile, osArchitecture, Charset.defaultCharset());
             }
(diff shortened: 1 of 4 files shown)


src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -109,20 +109,33 @@ public File installOrUpdateScanner() throws BlackDuckIntegrationException {
...
+
             // A version file exists, so we have to compare to determine if a download should occur.
             String localScannerVersion = FileUtils.readFileToString(versionFile, Charset.defaultCharset());
+            // An architecture file exists, so we will compare the architecture to determine if a download should occur
+            String localArchitecture = FileUtils.readFileToString(architectureFile, Charset.defaultCharset());
+
+            // Initiate a fresh download and update architecture value in file
+            if(!localArchitecture.equals(osArchitecture)) {
+                localScannerVersion = "";
+                FileUtils.writeStringToFile(architectureFile, osArchitecture, Charset.defaultCharset());
```
andrian-sevastyanov, line 136: It would be a bit safer to write this file AFTER the installation succeeds. Otherwise, we could end up with a mismatched architecture file.

PR #469, opened 2025-08-20
Add support for new metadata.json file in new scan-cli

devmehtabd:
This PR takes in a new metadata.json file which is packaged with new scan cli. This will handle the case when arm64 users will try to upgrade from 10.6.0 to 10.7.0 version of Detect and have locally cached scan cli. Currently there is no way for Detect to differentiate the architecture between old and new scan cli for us to be able to update the locally stored scan cli.

There are new conditions when fresh download will get triggered, if the metadata file does not exist and if it exists the current osArchitecture value is not equal to the one stored in file and also checks if the os value is different from the expected one.

### DIFF
--- a/src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
+++ b/src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
@@ -22,14 +22,12 @@ import com.blackduck.integration.rest.response.Response;
 import com.blackduck.integration.util.CleanupZipExpander;
 import com.blackduck.integration.util.OperatingSystemType;
 import org.apache.commons.compress.archivers.ArchiveException;
-import org.apache.commons.io.FileUtils;
 
 import javax.net.ssl.HttpsURLConnection;
 import javax.net.ssl.SSLHandshakeException;
 import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
-import java.nio.charset.Charset;
 import java.nio.file.Files;
 import java.nio.file.Paths;
 import java.security.cert.Certificate;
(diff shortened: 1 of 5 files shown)


src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -108,28 +106,50 @@ public File installOrUpdateScanner() throws BlackDuckIntegrationException {
...
         try {
-            String scannerVersion;
-            if (!versionFile.exists()) {
-                logger.info("No version file exists, assuming this is new installation and the signature scanner should be downloaded.");
-                scannerVersion = downloadSignatureScanner(scannerExpansionDirectory, downloadUrl, "");
-                // Version file creation should happen after successful download
-                logger.debug("The version file has not been created yet so creating it now.");
-                FileUtils.writeStringToFile(versionFile, scannerVersion, Charset.defaultCharset());
+            ScanCliMetadata scanCliMetadata;
+            if (scanCliMetadataFile == null || !scanCliMetadataFile.exists()) {
```
andrian-sevastyanov, line 115: Until customers upgrade their BD to 2025.7+, this will cause downloading the ZIP every time. Perhaps, if the file doesn't exist we should assume the default arch?

src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -108,28 +106,50 @@ public File installOrUpdateScanner() throws BlackDuckIntegrationException {
...
-            scannerVersion = downloadSignatureScanner(scannerExpansionDirectory, downloadUrl, localScannerVersion);
-            // Update version file if needed
-            if (localScannerVersion != scannerVersion) {
-                FileUtils.writeStringToFile(versionFile, scannerVersion, Charset.defaultCharset());
+
+            scanCliMetadata = ScanCliMetadata.getMetadata(scanCliMetadataFile);
+            // determine last version from the file
+            String localScannerVersion = scanCliMetadata.getToolVersion();
+            
+            String localArchitecture = scanCliMetadata.getArch();
```
andrian-sevastyanov, line 139: Should we check the OS as well?

src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -108,28 +106,50 @@ public File installOrUpdateScanner() throws BlackDuckIntegrationException {
...
+            if(localArchitecture.equals("arm64")) {
```
andrian-sevastyanov, line 140: minor: would be good to have constants for all these strings we're comparing.

src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
devmehtabd: It will do it anyways, since we now do not have the version info unless we do not want to completely delete the version file from the directory?

src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
devmehtabd: Sure, doesn't hurt to check it

src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
devmehtabd: I did try to cover all the scenarios, now we will have both version file and metadata file. We will remove the version file once we end support for Detect 10.6.0 and below.

psantos1113 (approved):

dterrybd (approved):

src/main/java/com/blackduck/integration/blackduck/codelocation/signaturescanner/command/ToolsApiScannerInstaller.java
```diff
@@ -106,50 +112,50 @@ public File installOrUpdateScanner() throws BlackDuckIntegrationException {
         File scannerExpansionDirectory = new File(installDirectory, BLACK_DUCK_SIGNATURE_SCANNER_INSTALL_DIRECTORY);
         scannerExpansionDirectory.mkdirs();
 
+        versionFile = new File(scannerExpansionDirectory, VERSION_FILENAME);
```
andrian-sevastyanov, line 115: nit: Ideally this should always be initialized or not be an instance variable (passed around to all methods that need it as an argument, similar to `scannerExpansionDirectory`). All other instance variables in this class are final.

